PR #3, opened 2019-03-05
Add a couple of lvsfunc functions

LightArrowsEXE:
- Added getw (originally from kagefunc)
- Added is_image
- Added source

_Source_, while a pretty general thing, might not be worth putting on here. lmk if it isn't and I'll remove it.

### DIFF
--- a/vsutil.py
+++ b/vsutil.py
@@ -3,6 +3,7 @@ VSUtil. A collection of general-purpose Vapoursynth functions to be reused in mo
 """
 from functools import reduce
 import vapoursynth as vs
+import mimetypes                    # added by stux!
 import re
 
 core = vs.core
--- a/vsutil.py
+++ b/vsutil.py
@@ -89,7 +89,7 @@ def join(planes: list, family=vs.YUV) -> vs.VideoNode:
     return core.std.ShufflePlanes(clips=planes, planes=[0], colorfamily=family)
 
 
-def getw(h, ar=16 / 9, only_even=True):
+def getw(h: int, ar: float=16/9, only_even: bool=True) -> int:
     """
     returns width for image (taken from kagefunc)
     """
--- a/vsutil.py
+++ b/vsutil.py
@@ -3,8 +3,7 @@ VSUtil. A collection of general-purpose Vapoursynth functions to be reused in mo
 """
 from functools import reduce
 import vapoursynth as vs
-import mimetypes                    # added by stux!
-import re
+import mimetypes
 
 core = vs.core
 
--- a/vsutil.py
+++ b/vsutil.py
@@ -103,7 +103,7 @@ def is_image(filename: str) -> bool:
     """
     Returns true if a filename refers to an image.
     """
-    return bool(re.search(r'\.(png|jpe?g|bmp|tiff?)$', filename))
+    return mimetypes.types_map[os.path.splitext(filename)[-1]].startswith("image/")
 
 
 def source(file: str, force_lsmas=False) -> vs.VideoNode:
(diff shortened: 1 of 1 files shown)


vsutil.py
```diff
@@ -85,3 +86,44 @@ def join(planes: list, family=vs.YUV) -> vs.VideoNode:
...
+    if only_even:
+        w = w // 2 * 2
+    return w
+
+
+def is_image(filename: str) -> bool:
+    """
+    Returns true if a filename refers to an image.
+    """
+    return bool(re.search(r'\.(png|jpe?g|bmp|tiff?)$', filename))
```
stuxcrystal: https://docs.python.org/3/library/os.path.html#os.path.splitext
```suggestion
    return os.path.splitext(filename)[-1] in {"png", "jpg", "jpeg", "bmp", "tif", "tiff"}
```

vsutil.py
```diff
@@ -85,3 +86,44 @@ def join(planes: list, family=vs.YUV) -> vs.VideoNode:
...
+
+
+def source(file: str, force_lsmas=False) -> vs.VideoNode:
+    """
+    Quick general import wrapper that automatically matches various sources with an appropriate indexing filter.
+    """
+    if file.startswith("file:///"):
+        file = file[8::]
+
+    if force_lsmas:
```
stuxcrystal:
```suggestion
    if force_lsmas:
        return core.lsmas.LWLibavSource(file)
        
    if file.endswith(".d2v"):
        clip = core.d2v.Source(file)
    elif is_image(file):
        clip = core.imwri.Read(file)
    elif file.endswith(".m2ts"):
        clip = core.lsmas.LWLibavSource(file)
    else:
        clip = core.ffms2.Source(file)

    return clip 
```

vsutil.py
stuxcrystal: Or what about letting the OS tell you if it is an image.

```suggestion
    return mimetypes.types_map[os.path.splitext(filename)[-1]].startswith("image/")
```

vsutil.py
stuxcrystal: Now read the code again, check if would work and be embarrased.

PS: The suggestion-features does not support multi-line suggestions.

vsutil.py
```diff
@@ -103,7 +103,7 @@ def is_image(filename: str) -> bool:
     """
     Returns true if a filename refers to an image.
     """
-    return bool(re.search(r'\.(png|jpe?g|bmp|tiff?)$', filename))
+    return mimetypes.types_map[os.path.splitext(filename)[-1]].startswith("image/")
```
stuxcrystal: 🤔 I think I forgot to add an import.

```python
import mimetypes
```

stuxcrystal: Did that for you.

LightArrowsEXE: Yeah, it also appears to cause issues with anything that isn't an image when trying to use it in my own func now. Evidently I did not test this enough. :feelsdumbman: Looking into it atm

stuxcrystal: This should fix it. Since `types_map` is apparently a dict. So we use `dict.get(k, d)`

```suggestion
    return mimetypes.types_map.get(os.path.splitext(filename)[-1], "").startswith("image/")
```

LightArrowsEXE: Tested it (alongside double-checking if it works as intended while I was it), and it works. Thanks a stack!

LightArrowsEXE:
Yeah, I'll update it in lvsfunc and push it later. Thanks a lot 👍

LightArrowsEXE:
Actually I misunderstood your comment. 🤦‍♂️  I don't know exactly how those type annotations are used, so could you add them for me?

LightArrowsEXE:
Oh, I thought it'd be more complicated than that. Thanks!

stuxcrystal:
I can see why

```py
def iterate(base: T, function: Callable[[T], T], count: int) -> T:
```

might look hard. 

PS: That's from my latest PR.

LightArrowsEXE:
Yeah, that's why I thought it'd be better to leave it to you haha

kageru:
I originally didn’t include get_w, but if there’s actual demand, I guess I’ll merge it.

As for the rest:
@LightArrowsEXE 
1. I don’t want the source function. It doesn’t fit the scope of what vsutil is supposed to be.
2. The re import should be unused in the latest commit, so please remove that before we merge it.
@stuxcrystal 
1. Please remove the `#added by stux` from the mimetypes import. I don’t want to spoil the fun, but I don’t think it adds anything. 
2. Do we need type annotations for parameters that have defaults? An IDE should realize that 16/9 is a float.

stuxcrystal:
About 4. I consider it good form. The type of default may not encompass the entire range of values.

kageru:
I think github’s markdown was a little to smart for me there. I updated the last comment, and it should now make clear who should have addressed which issue. Just in case someone wants to read it in the future. (◕‿◕✿)

PR #4, opened 2019-03-06
Add ExprStr

Endilll:
Drop-in wrapper for Expr() expressions in infix form, because postfix form is hard for both writing and reading, therefore error-prone. Docstring explains the rest.

I'm aware that grammar can be defined in more generalized way, but I decided to keep things simple.

I wrote it for [BeatriceTools](https://github.com/Beatrice-Raws/BeatriceTools), but I think it fits nicely here.

### DIFF
--- a/vsutil.py
+++ b/vsutil.py
@@ -1,7 +1,11 @@
 """
 VSUtil. A collection of general-purpose Vapoursynth functions to be reused in modules and scripts.
 """
+import ast
 from functools import reduce
+import re
+from typing import Any, List
+
 import vapoursynth as vs
 
 core = vs.core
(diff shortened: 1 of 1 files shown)


kageru:
I understand your reasoning, but I think the sheer size of this method justifies it being in a separate module. I would also like to keep Exprs in vsutil in RPN, as that is the standard way for Vapoursynth. Everything else would result in a mix depending on personal preference, and I want to avoid that.

I know that @stuxcrystal worked on something similar for yuuno, so he might be interested in this code to add it as a default function to the editor (if he doesn’t already have a working implementation), but I’ll decline the PR here, because a 150 loc AST parser and RPN converter is beyond the intended scope.

PR #5, opened 2019-03-06
Add some type annotations

stuxcrystal:


kageru:
`fallback` should be `value: Optional[T], fallback_value: T) -> T`. That’s how I intend for it to be used, and I guess we should make that clear through type annotations.
Also don’t forget to remove the typevar for `V`, then.

Also, good to know Python supports type hints like these. I’ll use them myself in the future. Thank you.

stuxcrystal:
done.

PR #12, opened 2019-03-08
Add frame2clip from Yuuno Core  to vsutil

stuxcrystal:
From https://github.com/Irrational-Encoding-Wizardry/yuuno-core/blob/dbfc65c21a6d6a0f8a130ed3242abe16cc0fa8f1/yuuno/vs/clip.py#L198


kageru:
What is the use case of this for regular encoding? Encoders very rarely use VideoFrames in my experience.

kageru:
So I guess we can close this? @stuxcrystal

kageru:
After a discussion that happened on discord, we decided to merge this after all.

PR #14, opened 2019-04-10
Add helper for MT in scripts.

stuxcrystal:
Will help speeding up blocking filter instantiations.


kageru:
While this might have been very useful for projects like yuuno, I don’t think it is really applicable to regular Vapoursynth scripts. Very little time is spent on interpreting the actual script, so adding threading to that offers almost no benefit while still introducing room for error.

PR #19, opened 2019-04-10
Update gitignore to exclude pycharm config files

kageru:


stuxcrystal:
Did you try https://gitignore.io

kageru:
I have, but that assumes that all developers use PyCharm and share some of the config. I don’t .idea in the repo at all. It doesn’t belong here.

stuxcrystal:
https://gitignore.io/api/pycharm+all

kageru:
There is no additional value to adding all subdirectories of .idea manually (like gitignore.io does) if we’re excluding the entirety of .idea anyway. It just adds more clutter to the gitignore.

FichteFoll:
FWIW, imo the proper procedure to ignore your IDE or editing tool's meta data is to ignore them in a [global gitignore](https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore).

PR #20, opened 2019-04-10
Added missing tests

kageru:
That should bring us to 100% coverage

### DIFF
--- a/tests/test_vsutil.py
+++ b/tests/test_vsutil.py
@@ -12,8 +12,10 @@ class VsUtilTests(unittest.TestCase):
     YUV411P8_CLIP = vs.core.std.BlankClip(_format=vs.YUV411P8, width=160, height=120, color=[0, 128, 128], length=100)
     YUV440P8_CLIP = vs.core.std.BlankClip(_format=vs.YUV440P8, width=160, height=120, color=[0, 128, 128], length=100)
 
-    BLACK_SAMPLE_CLIP = vs.core.std.BlankClip(_format=vs.YUV420P8, width=160, height=120, color=[0, 128, 128], length=100)
-    WHITE_SAMPLE_CLIP = vs.core.std.BlankClip(_format=vs.YUV420P8, width=160, height=120, color=[255, 128, 128], length=100)
+    BLACK_SAMPLE_CLIP = vs.core.std.BlankClip(_format=vs.YUV420P8, width=160, height=120, color=[0, 128, 128],
+                                              length=100)
+    WHITE_SAMPLE_CLIP = vs.core.std.BlankClip(_format=vs.YUV420P8, width=160, height=120, color=[255, 128, 128],
+                                              length=100)
 
     def assert_same_dimensions(self, clip_a: vs.VideoNode, clip_b: vs.VideoNode):
         """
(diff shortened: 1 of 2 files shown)


tests/test_vsutil.py
```diff
@@ -62,6 +66,14 @@ def test_subsampling(self):
         self.assertEqual('422', vsutil.get_subsampling(self.YUV422P8_CLIP))
         self.assertEqual('411', vsutil.get_subsampling(self.YUV411P8_CLIP))
         self.assertEqual('410', vsutil.get_subsampling(self.YUV410P8_CLIP))
+        # let’s create a custom format with higher subsampling than any of the legal ones to test that branch as well:
+        try:
```
stuxcrystal: https://docs.python.org/dev/library/unittest.html#unittest.TestCase.assertRaises

tests/test_vsutil.py
```diff
@@ -70,6 +82,19 @@ def test_depth(self):
     def test_plane_size(self):
         self.assertEqual((160, 120), vsutil.get_plane_size(self.YUV420P8_CLIP, 0))
         self.assertEqual((80, 60), vsutil.get_plane_size(self.YUV420P8_CLIP, 1))
+        # these should fail because they don’t have a constant format or size
+        try:
```
stuxcrystal: https://docs.python.org/dev/library/unittest.html#unittest.TestCase.assertRaises

tests/test_vsutil.py
```diff
@@ -70,6 +82,19 @@ def test_depth(self):
...
+            vsutil.get_plane_size(
+                vs.core.std.Splice([self.BLACK_SAMPLE_CLIP, self.SMALLER_SAMPLE_CLIP], mismatch=True), 0)
+            self.fail("This should have failed.")
+        except ValueError:
+            pass
+        try:
```
stuxcrystal: https://docs.python.org/dev/library/unittest.html#unittest.TestCase.assertRaises

tests/test_vsutil.py
```diff
@@ -90,6 +115,13 @@ def test_insert_clip(self):
         self.assert_same_metadata(self.BLACK_SAMPLE_CLIP, inserted_middle)
         self.assert_same_metadata(self.BLACK_SAMPLE_CLIP, inserted_end)
 
+        # this should fail
+        try:
```
stuxcrystal: https://docs.python.org/dev/library/unittest.html#unittest.TestCase.assertRaises

tests/test_vsutil.py
```diff
@@ -110,3 +142,28 @@ def test_frame2clip(self):
         frame = self.WHITE_SAMPLE_CLIP.get_frame(0)
         clip = vsutil.frame2clip(frame)
         self.assert_same_frame(self.WHITE_SAMPLE_CLIP, clip)
+        # specifically test the path with disabled cache
+        vs.core.add_cache = False
```
stuxcrystal: Add try-finally, so this value is always reset.

tests/test_vsutil.py
```diff
@@ -110,3 +142,28 @@ def test_frame2clip(self):
...
+        self.assertEqual(vsutil.is_image('something.m2ts'), False)
+
+    def test_get_w(self):
+        self.assertEqual(vsutil.get_w(480, only_even=True), 854)
+        self.assertEqual(vsutil.get_w(480, only_even=False), 853)
+        self.assertEqual(vsutil.get_w(1080, 4 / 3), 1440)
+        self.assertEqual(vsutil.get_w(1080, 16 / 9), 1920)
+
+    def test_iterate(self):
+        def double_number(x: int) -> int:
```
stuxcrystal: Oh and it should also check if the function is called the correct amount of times.

TLDR: Code coverage isn't everything

kageru: But the result basically does that. If `2` is doubled `3` times, the result is `16`.

PR #26, opened 2019-09-09
Add clip_sum

kageru:
Figured this might be useful, at least I’ve seen a few situations where `sum` on clips would have been needed.

### DIFF
--- a/vsutil.py
+++ b/vsutil.py
@@ -166,3 +166,12 @@ def is_image(filename: str) -> bool:
     Returns true if a filename refers to an image.
     """
     return mimetypes.types_map.get(os.path.splitext(filename)[-1], "").startswith("image/")
+
+
+def clip_sum(*clips: List[vs.VideoNode]) -> vs.VideoNode:
+    """
+    Concatenates all input clips.
+    """
+    if len(clips) == 1:
+        return clips[0]
+    return sum(clips[1:], clips[0])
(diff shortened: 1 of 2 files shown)


vsutil.py
```diff
@@ -166,3 +166,12 @@ def is_image(filename: str) -> bool:
     Returns true if a filename refers to an image.
     """
     return mimetypes.types_map.get(os.path.splitext(filename)[-1], "").startswith("image/")
+
+
+def clip_sum(*clips: List[vs.VideoNode]) -> vs.VideoNode:
```
FichteFoll: Why not use an iterable like `sum` itself? That makes their signatures almost equivalent (except for `start`, which is the sole reason you added this anyway). Code will get a bit more complex as a result, however.

kageru: Isn’t this more user-friendly? I like the varargs here because it allows you to write `clip_sum(clipa, clipb, clipc)` rather than `clip_sum([clipa, clipb, clipc])`.

FichteFoll: Depends on what you are going for. I'd prefer signature equivalence with the `sum` function it is based on over saving two characters. (And it's actually more powerful, albeit seldom relevant.)

vsutil.py
```diff
@@ -166,3 +166,12 @@ def is_image(filename: str) -> bool:
...
+    """
+    Concatenates all input clips.
+    """
+    if len(clips) == 1:
+        return clips[0]
+    return sum(clips[1:], clips[0])
```
stuxcrystal: `return core.std.Splice(clips)`

which is faster than `clip_a+clip_b+clip_c+...` as this evaluates to `Splice([Splice(Splice([a, b]), c], ...)`

kageru: I completely forgot std.Splice takes varargs. In that case, we can just drop this PR altogether since it adds basically no value.

FichteFoll: And for completeness' sake, `core.std.Splice(*iterable)` is the iterable version of it.

PR #27, opened 2019-10-19
Add is_luma_only

LightArrowsEXE:
Adds a function to check if there is only a luma plane available in a given clip. Returns True or False. Useful for functions that handle just the luma and want to check what it has to return later (be it a GRAY or a full YUV clip). Better name suggestions are welcome.

### DIFF
--- a/vsutil.py
+++ b/vsutil.py
@@ -166,3 +166,10 @@ def is_image(filename: str) -> bool:
     Returns true if a filename refers to an image.
     """
     return mimetypes.types_map.get(os.path.splitext(filename)[-1], "").startswith("image/")
+
+
+def is_luma_only(clip: vs.VideoNode) -> Bool:
+    """
+    Returns true if a clip has only one plane
+    """
+    return format.num_planes == 1
--- a/vsutil.py
+++ b/vsutil.py
@@ -172,4 +172,4 @@ def is_luma_only(clip: vs.VideoNode) -> Bool:
     """
     Returns true if a clip has only one plane
     """
-    return format.num_planes == 1
+    return clip.format.num_planes == 1


vsutil.py
```diff
@@ -166,3 +166,10 @@ def is_image(filename: str) -> bool:
     Returns true if a filename refers to an image.
     """
     return mimetypes.types_map.get(os.path.splitext(filename)[-1], "").startswith("image/")
+
+
+def is_luma_only(clip: vs.VideoNode) -> Bool:
+    """
+    Returns true if a clip has only one plane
+    """
+    return format.num_planes == 1
```
FichteFoll: Did you mean `clip.format.num_planes`?

Either way, this is a function for simple comparison I don't personally feel worth adding a function for. Besides, you only actually test for the number of planes, not whether the plane is luma, so writing the actual expression in code is more explicit and easier to understand in a context.

LightArrowsEXE: Yeah, fixed that.

I think it is general enough to be worthwhile to have in vsutil, but I do think the name needs to be changed. That's why I asked for name suggestions in the original post.

FichteFoll:
Okay, here is my thought process:

What does the function do? Check if a clip is only luma? No. Check if a clip has only 1 plane? Yes. What would we name a function like that? `has_one_plane(clip)`. What expresses exactly the same semantics? `clip.format.num_planes == 1`. What is more explicit and barely has more characters? The latter. Thus you don't need this function.

LightArrowsEXE:
I still believe it to be worthwhile to include it, as a form of boilerplate. It's much easier to write a function and use the name of said function than it is to rewrite the same line every time, and considering a lot of functions either process just the luma or have an option to process just the luma, I believe this can still give value as a general utility function in vsutil.

kageru:
I agree with @FichteFoll. This doesn’t save any typing, and the name would either be inaccurate or offer no benefit over `clip.format.num_planes == 1`.
Vapoursynth has no way of differentiating between an isolated luma plane and any other grey clip. I don’t want to create the impression that it does by having a function like this.

PR #25, opened 2019-09-02
Check whether get_y is passed a yuv-clip

stuxcrystal:

PR #28, opened 2020-04-28
Some fixes, re-wording, better testing.

OrangeChannel:
More information in the commit descriptions.

Not sure why GitHub displays a huge remove/add chunk on tests L7-L17 since all I did was replace the weird `_format` arg with `format` in 9557a86.

EDIT: This PR closes #25 by the way.

### DIFF
--- a/vsutil.py
+++ b/vsutil.py
@@ -15,12 +15,13 @@ core = vs.core
 T = TypeVar('T')
 
 
-def get_subsampling(clip: vs.VideoNode) -> str:
+def get_subsampling(clip: vs.VideoNode) -> Union[None, str]:
     """
     Returns the subsampling of a VideoNode in human-readable format.
+    Returns None for formats without subsampling.
     """
     if clip.format.color_family not in (vs.YUV, vs.YCOCG):
-        return ''
+        return None
     if clip.format.subsampling_w == 1 and clip.format.subsampling_h == 1:
         return '420'
     elif clip.format.subsampling_w == 1 and clip.format.subsampling_h == 0:
--- a/vsutil.py
+++ b/vsutil.py
@@ -16,21 +16,22 @@ def get_subsampling(clip: vs.VideoNode) -> str:
     """
     Returns the subsampling of a VideoNode in human-readable format.
     """
+    if clip.format.color_family not in (vs.YUV, vs.YCOCG):
+        return ''
     if clip.format.subsampling_w == 1 and clip.format.subsampling_h == 1:
-        css = '420'
+        return '420'
     elif clip.format.subsampling_w == 1 and clip.format.subsampling_h == 0:
-        css = '422'
+        return '422'
     elif clip.format.subsampling_w == 0 and clip.format.subsampling_h == 0:
-        css = '444'
+        return '444'
     elif clip.format.subsampling_w == 2 and clip.format.subsampling_h == 2:
-        css = '410'
+        return '410'
     elif clip.format.subsampling_w == 2 and clip.format.subsampling_h == 0:
-        css = '411'
+        return '411'
     elif clip.format.subsampling_w == 0 and clip.format.subsampling_h == 1:
-        css = '440'
+        return '440'
     else:
-        raise ValueError('Unknown subsampling')
-    return css
+        raise ValueError('Unknown subsampling.')
 
 
 def get_depth(clip: vs.VideoNode) -> int:
--- a/vsutil.py
+++ b/vsutil.py
@@ -126,7 +126,8 @@ def split(clip: vs.VideoNode, /) -> List[vs.VideoNode]:
     return [plane(clip, x) for x in range(clip.format.num_planes)]
 
 
-def join(planes: List[vs.VideoNode], family: Literal[vs.RGB, vs.YUV, vs.YCOCG] = vs.YUV) -> vs.VideoNode:
+def join(planes: List[vs.VideoNode],
+         family: Literal[vs.ColorFamily.RGB, vs.ColorFamily.YUV, vs.ColorFamily.YCOCG] = vs.YUV) -> vs.VideoNode:
     """
     Joins the supplied list of planes into a three-plane VideoNode (defaults to YUV).
     """
--- a/vsutil.py
+++ b/vsutil.py
@@ -107,7 +107,7 @@ def get_y(clip: vs.VideoNode) -> vs.VideoNode:
     """
     Helper to get the luma of a VideoNode.
     """
-    if clip.format is None or clip.format.color_family != vs.YUV:
+    if clip.format is None or clip.format.color_family not in (vs.YUV, vs.YCOCG):
         raise ValueError("You did not pass a YUV-Clip into get_y.")
         
     return plane(clip, 0)
--- a/vsutil.py
+++ b/vsutil.py
@@ -12,7 +12,7 @@ from typing import Callable, List, Optional, Tuple, TypeVar, Union
 import vapoursynth as vs
 
 core = vs.core
-T = TypeVar("T")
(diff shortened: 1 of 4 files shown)


vsutil.py
```diff
@@ -146,7 +151,7 @@ def frame2clip(frame: vs.VideoFrame, *, enforce_cache=True) -> vs.VideoNode:
     # Forcefully add a cache to Modify-Frame if caching is disabled on the core.
     # This will ensure that the filter will not include GIL characteristics.
     if not core.add_cache and enforce_cache:
-        result = result.std.Cache(size=1, fixed=True)
+        return result.std.Cache(size=1, fixed=True)
```
stuxcrystal, line 154: I don't like this change:

If you want to add a feature afterwards, you need to remove the `return` and replace it back with `result = `

I suggest keeping the `result = `

vsutil.py
```diff
@@ -157,12 +162,12 @@ def get_w(height: int, aspect_ratio: float = 16 / 9, only_even: bool = True) ->
     """
     width = height * aspect_ratio
     if only_even:
-        width = round(width / 2) * 2
-    return int(round(width))
+        return round(width / 2) * 2
```
stuxcrystal, line 165: Same here.

FichteFoll: Sounds like YAGNI. I can understand the first comment because that is just the last step of a longer process, but here?

stuxcrystal: True.

OrangeChannel:
I'm going to force push some changes to keep the commit history clean(er). If github's annoying about it, here's the changes:

- I addressed <https://github.com/Irrational-Encoding-Wizardry/vsutil/pull/28#discussion_r417242424> and reverted the change (and changed the commit message as a result).
- Change `get_subsampling` output for non-YUV/YCoCG color family clips from a ValueError to an empty string. Might be more useful when doing `if get_subsampling(clip)`. Changed tests to reflect this change.
- Add an `__all__` dunder for those who will `from vsutil import *` in encoding scripts

vsutil.py
```diff
@@ -1,58 +1,63 @@
...
 T = TypeVar("T")
 
 
 def get_subsampling(clip: vs.VideoNode) -> str:
     """
-    Returns the subsampling of a clip in human-readable format.
+    Returns the subsampling of a VideoNode in human-readable format.
     """
+    if clip.format.color_family not in (vs.YUV, vs.YCOCG):
+        return ''
```
FichteFoll, line 23: I don't know about this. When would you be using this function on a clip without subsampling and would you expect an empty string as the return value? Since I don't know, I'm not blocking this, of course, but it does raise a question to me.

vsutil.py
```diff
@@ -1,58 +1,63 @@
...
     :return: (width, height)
     """
     # Add additional checks on Video Nodes as their size and format can be variable.
     if isinstance(frame, vs.VideoNode):
         if frame.width == 0:
-            raise ValueError("Cannot calculate plane size of variable size clip. Pass a frame instead.")
+            raise ValueError('Cannot calculate plane size of variable size clip. Pass a frame instead.')
         if frame.format is None:
-            raise ValueError("Cannot calculate plane size of variable format clip. Pass a frame instead.")
+            raise ValueError('Cannot calculate plane size of variable format clip. Pass a frame instead.')
```
FichteFoll, line 60: Why did you change quotation characters here? Do we have any guidelines on this? I see that both styles are still being used. 

(I know my guidelines, where I use single quotes for symbol-like identifiers and double for what could be described as text, but I am just a third party here.)

OrangeChannel: It was either we switched completely to `"` or `'` and I usually go with `'`, it's not a standard defined anywhere but either/or works, just didn't like seeing both.

vsutil.py
OrangeChannel: I was thinking if some function has two different ways of handling clips with/without subsampling, they could `if get_subsampling(clip): do something...` instead raising an error. And figured we wouldn't want two return types for such a simple function (a `False` here for example).

I also am going to later suggest a `get_color_family(clip: vs.VideoNode) -> str: return clip.format.color_family.name` function for vsutil that this function could make use of and maybe some other scripts as well.

vsutil.py
OrangeChannel: Now that you mention it actually, I missed the strings in `is_image()`, will update soon.

OrangeChannel: Well, it's documented with a single quote so I changed the TypeVar too <https://docs.python.org/3/library/typing.html#generics>

vsutil.py
FichteFoll: I want to suggest returning `None` in this case, because … well, it doesn't have any subsampling in that case. If people want to have that render as '', they can `f"{get_subsampling(clip) || ''}"`, for example.

OrangeChannel: Makes sense, but what should we reword `Returns the subsampling of a VideoNode in human-readable format.` to then? Or should we just leave it and just let the typehint `-> Union[None, str]` speak for itself.

FichteFoll: Typehint speaks for itself imo, but could add `Returns None for formats without subsampling.` regardless.

OrangeChannel: Used your suggestion in b408157, ready for merge now I believe.

```
============================= test session starts ==============================
platform linux -- Python 3.8.2, pytest-5.4.1, py-1.8.1, pluggy-0.13.1 -- /usr/bin/python3.8
cachedir: .pytest_cache
collecting ... collected 11 items

test_vsutil.py::VsUtilTests::test_depth PASSED                           [  9%]
test_vsutil.py::VsUtilTests::test_fallback PASSED                        [ 18%]
test_vsutil.py::VsUtilTests::test_frame2clip PASSED                      [ 27%]
test_vsutil.py::VsUtilTests::test_get_w PASSED                           [ 36%]
test_vsutil.py::VsUtilTests::test_get_y PASSED                           [ 45%]
test_vsutil.py::VsUtilTests::test_insert_clip PASSED                     [ 54%]
test_vsutil.py::VsUtilTests::test_is_image PASSED                        [ 63%]
test_vsutil.py::VsUtilTests::test_iterate PASSED                         [ 72%]
test_vsutil.py::VsUtilTests::test_plane_size PASSED                      [ 81%]
test_vsutil.py::VsUtilTests::test_split_join PASSED                      [ 90%]
test_vsutil.py::VsUtilTests::test_subsampling PASSED                     [100%]

============================== 11 passed in 0.10s ==============================
```

OrangeChannel:
Added some last minute fixes and an update to 3.8 syntax, since they make more sense to add in this PR than in the depth converter function PR. All tests passed.

vsutil.py
```diff
@@ -1,58 +1,64 @@
...
-import mimetypes
-import os
 
 core = vs.core
-T = TypeVar("T")
+T = TypeVar('T')
 
 
-def get_subsampling(clip: vs.VideoNode) -> str:
+def get_subsampling(clip: vs.VideoNode, /) -> Union[None, str]:
```
kageru, line 18: Why not use `Optional[str]` here?

vsutil.py
```diff
@@ -150,19 +160,19 @@ def frame2clip(frame: vs.VideoFrame, *, enforce_cache=True) -> vs.VideoNode:
     return result
 
 
-def get_w(height: int, aspect_ratio: float = 16 / 9, only_even: bool = True) -> int:
+def get_w(height: int, /, aspect_ratio: float = 16 / 9, *, only_even: bool = True) -> int:
```
kageru, line 163: I’d prefer not having height as positional-only here.
It feels like the kind of parameter to me that could be a named argument for clarity in some scripts, and I don’t want to break those.

PR #29, opened 2020-04-28
Add bit depth resample function.

OrangeChannel:
Based ontop of #28.


vsutil.py
```diff
@@ -157,12 +162,86 @@ def get_w(height: int, aspect_ratio: float = 16 / 9, only_even: bool = True) ->
...
+    :param bitdepth: Desired bits_per_sample of output clip.
+    :param force_float: Forces output clip to have FLOAT sample type,
+        only really useful for 16/32 bit conversions where the color family
+        supports both INTEGER and FLOAT at these bit depths.
+    :return: Converted clip with desired bit depth. ColorFamily will be same as input.
+    """
+
+    color_family = clip.format.color_family
+
+    if color_family == vs.GRAY and bitdepth not in (8, 16, 32):
```
stuxcrystal, line 189: Why those sanity checks with bitdepths?

vsutil.py
```diff
@@ -157,12 +162,86 @@ def get_w(height: int, aspect_ratio: float = 16 / 9, only_even: bool = True) ->
...
+        raise ValueError('Invalid bit depth for color family GRAY.')
+    if color_family == vs.YUV and bitdepth not in (8, 9, 10, 12, 14, 16, 32):
+        raise ValueError('Invalid bit depth for color family YUV.')
+    if color_family == vs.RGB and bitdepth not in (8, 9, 10, 16, 32):
+        raise ValueError('Invalid bit depth for color family RGB.')
+
+    # TODO: how to test COMPAT color family clips?
```
stuxcrystal, line 196: There are only two compat families. Test them manually if you have to.

stuxcrystal:
I'm gonna ask you to split your refactorings into a separate PR.

OrangeChannel:
Okay @stuxcrystal since I realized that the PresetFormats don't cover every possible format, and printing the name of the desired format in an error message was more of a pain in the ass than I cared for (what the hell is `YUV410PH16`, why is `RGB60` a valid format but `Core.resize` throws an exception when using it?), I've gotten rid of all the error checks and left the user to figure out VapourSynth's somewhat vague exception for themselves.

vsutil.py
```diff
@@ -157,12 +162,59 @@ def get_w(height: int, aspect_ratio: float = 16 / 9, only_even: bool = True) ->
...
 
 def is_image(filename: str) -> bool:
     """
     Returns true if a filename refers to an image.
     """
-    return mimetypes.types_map.get(os.path.splitext(filename)[-1], "").startswith("image/")
+    return types_map.get(path.splitext(filename)[-1], "").startswith("image/")
+
+
+def resample(clip: vs.VideoNode, bitdepth: int, *, force_float: bool = False) -> vs.VideoNode:
```
kageru, line 176: This function feels very verbose. Does it do more than just

```py
def depth(clip: vs.VideoNode, bitdepth: int, *, force_float: bool = False) -> vs.VideoNode:
    """
    A bit depth converter only using core.resize and the PresetFormats defined in vscore.cpp.
    By default, will always output INTEGER sample type.
    :param bitdepth: Desired bits_per_sample of output clip.
    :param force_float: Forces output clip to have FLOAT sample type,
        only really useful for 16/32 bit conversions where the color family
        supports both INTEGER and FLOAT at these bit depths.
    :return: Converted clip with desired bit depth. ColorFamily will be same as input.
    """
    curr_depth = get_depth(clip)
    target_sample_type = vs.FLOAT if force_float else vs.INTEGER
    # nothing to do
    if curr_depth == bitdepth and clip.format.sample_type = target_sample_type:
        return clip
    dither = 'error_diffusion' if curr_depth > bitdepth else dither = 'none'

    return clip.resize.Point(format=clip.format.replace(bits_per_sample=bitdepth, sample_type=target_sample_type), dither_type=dither)
```
?
I also prefer the name depth or something that very clearly communicates that this is just depth conversion.
And I think we should default to float for 32bit output because 32bit int is very rarely what a user wants or expects.

OrangeChannel: Was discussed on Discord: since switching either sample_type or bits_per_sample will add an overhead, even if one of them is the same as the source clip, it *can* be reduced to the following with no effect on the tests (or performance theoretically):
```py
def resample(clip: vs.VideoNode, bitdepth: int, *, sample_type: Optional[Literal[vs.FLOAT, vs.INTEGER]] = None) \
        -> vs.VideoNode:
    """
    A bit depth converter only using core.resize and Format.replace.
    By default, outputs FLOAT sample type for 32 bit and INTEGER for anything else.

    :param bitdepth:    Desired bits_per_sample of output clip.
    :param sample_type: Desired sample_type of output clip. Allows overwriting default FLOAT/INTEGER behavior.
    :return: Converted clip with desired bit depth and sample type. ColorFamily will be same as input.
    """
    curr_depth = get_depth(clip)
    if not sample_type:
        if bitdepth == 32:
            sample_type = vs.FLOAT
        else:
            sample_type = vs.INTEGER

    if curr_depth == bitdepth and clip.format.sample_type == sample_type:
        return clip

    dither = 'error_diffusion' if curr_depth > bitdepth and sample_type == vs.INTEGER else 'none'

    return clip.resize.Point(format=clip.format.replace(bits_per_sample=bitdepth, sample_type=sample_type),
                             dither_type=dither)
```

kageru: Please check for `if sample_type is None` explicitly. The sample type is an enum, and vs.INTEGER is 0, so
```py
>>> bool(vs.INTEGER)
False
```

OrangeChannel: Oh nice catch, fixed:
```py
def resample(clip: vs.VideoNode, bitdepth: int, *, sample_type: Optional[Literal[vs.FLOAT, vs.INTEGER]] = None) \
        -> vs.VideoNode:
    """
    A bit depth converter only using core.resize and Format.replace.
    By default, outputs FLOAT sample type for 32 bit and INTEGER for anything else.

    :param bitdepth:    Desired bits_per_sample of output clip.
    :param sample_type: Desired sample_type of output clip. Allows overriding default FLOAT/INTEGER behavior.
    :return: Converted clip with desired bit depth and sample type. ColorFamily will be same as input.
    """
    curr_depth = get_depth(clip)
    if sample_type is None:
        if bitdepth == 32:
            sample_type = vs.FLOAT
        else:
            sample_type = vs.INTEGER

    if curr_depth == bitdepth and clip.format.sample_type == sample_type:
        return clip

    dither = 'error_diffusion' if curr_depth > bitdepth and sample_type == vs.INTEGER else 'none'

    return clip.resize.Point(format=clip.format.replace(bits_per_sample=bitdepth, sample_type=sample_type),
                             dither_type=dither)
```

Frechdachs: - There should be a way to set the dither_type, in many situations you don't need dithering.

- Also, theoretically, dithering has to be done when upsampling full range content. (btw, zimg will not do dithering when upsampling limited range int to limited range int, even if specified.)

- When going float --> int there is currently no way to output full range.

- I would use a different name, since resampling can be much more than just bitdepth resampling.

OrangeChannel: Good point on the `dither_type` param, definitely should be changeable :+1: , not familiar enough with the zimg codebase nor vs to comment on the 2nd and 3rd point. Regarding the 4th point, the function name isn't set in stone yet, but what other resampling can be done besides changing the `sample_type` and the `bits_per_sample` as this currently lets you change both.

Frechdachs: >Regarding the 4th point, the function name isn't set in stone yet, but what other resampling can be done besides changing the sample_type and the bits_per_sample as this currently lets you change both.

e. g. size, subsampling, color format, chroma alignment
All of those can be changed with `fmtc.resample`.
`depth` might be an appropriate name, since you basically only want to call this function when you want to change the bitdepth. But maybe you have a better idea. There is also `placebo.Resample` which just lets you change the size. I'd find it counterintuitive for users if there are multiple "resample" functions with different features.

Regarding the second point. I think you can keep this default:
```python
    dither = 'error_diffusion' if curr_depth > bitdepth and sample_type == vs.INTEGER else 'none'
```
But I am not a 100% certain.
Strictly speaking, you need dithering when raising the bitdepth of full range input, or when converting between full and limited range. But it probably doesn't matter at high bitdepths and not using dithering is _much_ faster.

vsutil.py
```diff
@@ -176,3 +195,75 @@ def is_image(filename: str, /) -> bool:
...
+        try:
+            range_in = Range(range_in)
+        except ValueError:
+            raise ValueError(f'depth: range_in must be in {readable_enums(Range)}.') from None
+    if dither_type is not None:
+        try:
+            dither_type = Dither(dither_type)
+        except ValueError:
+            raise ValueError(f'depth: dither_type must be in {readable_enums(Dither)}.') from None
+
```
kageru, line 251: We only have to write this code once.
I’m thinking of something like
```py
def resolve_enum(enum_type, value):
    try:
        return enum_type(value)
    except ValueError:
        raise ValueError(f'value must be in {[f"{module}.{str(e)} ({e.value})" for e in enum_type]}') from None
```
which would look like
```py
In [22]: resolve_enum(Range, 0)
Out[22]: <Range.LIMITED: 0>

In [23]: resolve_enum(Range, 1)
Out[23]: <Range.FULL: 1>

In [24]: resolve_enum(Range, 2)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-24-7a2f94c717b0> in <module>
----> 1 resolve_enum(Range, 2)

<ipython-input-17-e5bc1e1b3f60> in resolve_enum(enum_type, value)
      3         return enum_type(value)
      4     except ValueError:
----> 5         raise ValueError(f'value must be in {[f"{module}.{str(e)} ({e.value})" for e in enum_type]}') from None
      6

ValueError: value must be in ['vsutil.Range.LIMITED (0)', 'vsutil.Range.FULL (1)']
```

vsutil.py
```diff
@@ -176,3 +195,75 @@ def is_image(filename: str, /) -> bool:
...
+    curr_depth = get_depth(clip)
+    if sample_type is None:
+        if bitdepth == 32:
+            sample_type = vs.FLOAT
+        else:
+            sample_type = vs.INTEGER
```
kageru, line 257:
```py
sample_type = fallback(sample_type, vs.FLOAT if bitdepth == 32 else vs.INTEGER)
```

vsutil.py
```diff
@@ -176,3 +195,75 @@ def is_image(filename: str, /) -> bool:
...
+
+    if (curr_depth, clip.format.sample_type, range_in) == (bitdepth, sample_type, range):
+        return clip
+
+    if dither_type is None:
+        if range_in != range or range_in == 1:  # thanks @Frechdachs for explaining this: 'you need dithering when raising
+            dither_type = Dither.ERROR_DIFFUSION  # the bitdepth of full range [or] converting between full and limited'
```
kageru, line 264: Please don’t use line end comments like this. These two form a coherent sentence, so either put them on one line or on two consecutive lines without code in-between.

vsutil.py
```diff
@@ -176,3 +195,75 @@ def is_image(filename: str, /) -> bool:
...
+        else:
+            dither_type = Dither.ERROR_DIFFUSION if curr_depth > bitdepth and sample_type == vs.INTEGER else Dither.NONE
```
kageru, line 266: There are only two possible outcomes here (none or error diffusion), but we have three code paths.
We could express it as a single if/else:
```py
if range_in != range or range_in == Range.FULL or (curr_depth > bitdepth and sample_type == vs.INTEGER):
  Dither.ERROR_DIFFUSION
else:
  Dither.NONE

File: vsutil.py [hunk 8]
```diff
@@ -176,3 +195,75 @@ def is_image(filename: str, /) -> bool:
...
+
+    return clip.resize.Point(format=clip.format.replace(bits_per_sample=bitdepth, sample_type=sample_type), range=range,
+                             range_in=range_in, dither_type=dither_type.value)
```
kageru, line 269: probably a personal thing, but if we move `range=range` to the second line, both lines have about the same length which I think looks nicer

tests/test_vsutil.py
```diff
@@ -161,3 +161,41 @@ def double_number(x: int) -> int:
 
         self.assertEqual(vsutil.iterate(2, double_number, 3), 16)
         self.assertEqual(vsutil.iterate(0, double_number, 4), 0)
+
+    def test_depth(self):  # TODO: test dither/range/range_in logic
```
kageru, line 165: TODO for whom? Will you do that in a future PR? Why not now?

vsutil.py
```diff
@@ -176,3 +195,75 @@ def is_image(filename: str, /) -> bool:
...
+        """
+        return [module + "." + str(i) for i in enum] + [i.value for i in enum]
+
+    # enum mapping and error checking
+    if sample_type is not None:
+        try:
+            sample_type = vs.SampleType(sample_type)
+        except ValueError:
+            raise ValueError(f'depth: sample_type must be in {readable_enums(vs.SampleType, str(vs.__name__))}.') from None
+    if range is not None:
```
kageru, line 236: What if it *is* `None`? Does core.resize handle that for us?

tests/test_vsutil.py
```diff
@@ -161,3 +161,41 @@ def double_number(x: int) -> int:
...
+        l_int_10_clip = limited_clip.resize.Point(format=limited_clip.format.replace(bits_per_sample=10))
+        l_int_16_clip = limited_clip.resize.Point(format=limited_clip.format.replace(bits_per_sample=16))
+        l_float_16_clip = limited_clip.resize.Point(format=limited_clip.format.replace(bits_per_sample=16, sample_type=vs.FLOAT))
+        l_float_32_clip = limited_clip.resize.Point(format=limited_clip.format.replace(bits_per_sample=32, sample_type=vs.FLOAT))
+
+        self.assertEqual(vsutil.depth(full_clip, 8), full_clip)
+        self.assert_same_format(vsutil.depth(full_clip, 10), int_10_clip)
+        self.assert_same_format(vsutil.depth(full_clip, 16), int_16_clip)
+        self.assert_same_format(vsutil.depth(full_clip, 16, sample_type=vs.FLOAT), float_16_clip)
+        self.assert_same_format(vsutil.depth(full_clip, 32), float_32_clip)
```
kageru, line 191: As far as I know, limited/full range is not part of the format, so this assertion won’t check for that. If that’s known/expected, then it’s fine. Just wanted to point it out.

vsutil.py
```diff
@@ -176,3 +195,75 @@ def is_image(filename: str, /) -> bool:
...
+        Defaults to Floyd-Steinberg error diffusion when downsampling, converting between ranges, or upsampling full
+        range input. Defaults to 'none', or round to nearest, otherwise.
+
+    :return: Converted clip with desired bit depth and sample type. ColorFamily will be same as input.
+    """
+    def readable_enums(enum, module: str = 'vsutil') -> list:
+        """
+        Returns a readable error message listing all possible values in `module.enum`.
+        """
+        return [module + "." + str(i) for i in enum] + [i.value for i in enum]
```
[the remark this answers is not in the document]
FichteFoll: For the record, an enum's `__repr__` would do it like this:
```py
['<Range.LIMITED: 0>', '<Range.FULL: 1>']
```

The only reason for not using repr directly here must be because one of the enums is from the vs module and it would thus be misleading where that could come from. Otoh, `vsutil` could reexport that enum (and list it in `__all__`).

vsutil.py
FichteFoll: If accessing generic vars from types wasn't such a hassle, I would suggest getting the type from the function's `__annotations__` even, but this would be less readable.

vsutil.py
FichteFoll:
```py 
dither_type = (dither_type 
               or (Dither.ERROR_DIFFUSION
                   if (range_in != range 
                       or range_in == Range.FULL 
                       or (curr_depth > bitdepth and sample_type == vs.INTEGER))
                   else Dither.NONE))
```
yeah, I'm flexing

vsutil.py
FichteFoll: My take is to have each parameter on its own line. LoC is a stupid metric.

vsutil.py
kageru: I thought about that as well, but in the end it almost felt less readable to me, despite probably being more pythonic.
EIther way, if you’re going with that, you can use `fallback(dither_type, ...)` instead of `(dither_type or ...)`.

FichteFoll:
```py
should_dither = (range_in != range
                 or range_in == Range.FULL
                 or (curr_depth > bitdepth and sample_type == vs.INTEGER)
dither_type = fallback(dither_type, Dither.ERROR_DIFFUSION if should_dither else Dither.NONE)
```
this is less flexy

vsutil.py
stuxcrystal: Why not manually override `__repr__`?

Something along the lines of:

```py
    def __repr__(self):
        return f"<{self.__module__}.{self.__qualname__}.{self.name}: {self.value}>"
```

vsutil.py
```diff
@@ -176,3 +195,75 @@ def is_image(filename: str, /) -> bool:
...
+                        vapoursynth.SampleType enums INTEGER and FLOAT or their values, [0, 1].
+    :param range:       Output pixel range (defaults to input clip's range). See `Range`.
+    :param range_in:    Input pixel range (defaults to input clip's range). See `Range`.
+    :param dither_type: Dithering algorithm. Allows overriding default dithering behavior. See `Dither`.
+        Defaults to Floyd-Steinberg error diffusion when downsampling, converting between ranges, or upsampling full
+        range input. Defaults to 'none', or round to nearest, otherwise.
+
+    :return: Converted clip with desired bit depth and sample type. ColorFamily will be same as input.
+    """
+    def readable_enums(enum, module: str = 'vsutil') -> list:
```
stuxcrystal, line 224: I personally think this function might come handy in future functions, what about moving the function up to module level, perhaps mark it as private by prepending a `_` to the function name?

vsutil.py
```diff
@@ -1,9 +1,10 @@
 """
 VSUtil. A collection of general-purpose VapourSynth functions to be reused in modules and scripts.
 """
-__all__ = ['core', 'fallback', 'frame2clip', 'get_depth', 'get_plane_size', 'get_subsampling', 'get_w', 'get_y',
-           'insert_clip', 'is_image', 'iterate', 'join', 'plane', 'split', 'vs']
+__all__ = ['Dither', 'Range', 'core', 'depth', 'fallback', 'frame2clip', 'get_depth', 'get_plane_size',
```
stuxcrystal, line 4: Why do you export `core`? Actually, why did we ever export `core`?

FichteFoll: FWIW, `vs` is also exported.

tests/test_vsutil.py
OrangeChannel: The testing right now would require either a debug parameter in the function, or by using a hacky method with `sys.argv` and checking how the script was called (i.e. with pytest or unittest). As discussed on Discord, the hacky method is ugly and is only needed if we're trying to go for 100% code coverage again, but considering we don't even have a working CI testing system right now, there's really no need to in-depth test as I'm only using it here to make sure the function actually returns the expected clip/format. Another thing that might be better placed in another PR is testing this on actual example frames/clips to make sure the resize call (specifically the dithering) is working properly. I'd rather not bloat the repo with images though and would want to reliably host them somewhere else.

I'm also not even sure of a good way to test the range logic yet, so to answer `for whom` for this part, I guess whoever can find a good way to test this.

vsutil.py
OrangeChannel: From the original commit, <https://github.com/Irrational-Encoding-Wizardry/vsutil/pull/28/commits/45d22b76caa03bd3b785fe22a0cbd611ceb79246>
> This allows cleaner looking scripts, and might as well be the
recommended way of import from vsutil _in scripts_ not other modules.
With `core` and `vs` being defined in __all__, a VS script will not need
both of the standard VS lines if using the wild card import.

In someone's encoding script, a `from vsutil import *` will allow them to access all `vs` members and use the normal `core.Etc.etc` syntax without having to write:

```py
import vapoursynth as vs
from vsutil import *
core = vs.core
```
at the top of every script.

vsutil.py
OrangeChannel: Sounds good to me.

vsutil.py
OrangeChannel: Gonna link the Discord discussion here: <https://discordapp.com/channels/221919789017202688/224104675031384065/715383522613002272>
This is a really good question that I'm not exactly sure has ever been handled by fvf.Depth correctly either.

vsutil.py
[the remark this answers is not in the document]
OrangeChannel: @FichteFoll getting the module name from `enum.__module__` doesn't work with imported modules.

```py
print(_readable_enums(vs.ColorFamily))

['importlib._bootstrap.ColorFamily.GRAY (1000000)', 'importlib._bootstrap.ColorFamily.RGB (2000000)', 'importlib._bootstrap.ColorFamily.YUV (3000000)', 'importlib._bootstrap.ColorFamily.YCOCG (4000000)', 'importlib._bootstrap.ColorFamily.COMPAT (9000000)']
```

Even calling it on `Range` gives `['__main__.Range.LIMITED (0)', '__main__.Range.FULL (1)']`

vsutil.py
OrangeChannel: Addressed in d78e2bace67a92d2ac48bc13216245a888505a72.

vsutil.py
OrangeChannel: Addressed in a9c672116db536ada8c16c06843b89932266a234.

vsutil.py
OrangeChannel: Addressed in a9c672116db536ada8c16c06843b89932266a234.

vsutil.py
OrangeChannel: Addressed in 684a0415ba6a7e85f7606279ea55bc17009ba3a1.

vsutil.py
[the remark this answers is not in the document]
OrangeChannel: Took FichteFoll's suggestion to extend the default `repr(enum.value)` behavior to prepend module name to the enum in 684a0415ba6a7e85f7606279ea55bc17009ba3a1.

vsutil.py
OrangeChannel: Took Fichte's suggestion in d4047637be705c81c577a2edd49e631ab0bfb6d5.

tests/test_vsutil.py
OrangeChannel: Yea, just wanted to test both RGB and YUV clips here really, and that was just how I named them. Not sure if it's actually necessary to check both anyways, so might be better to just check on the full_clip.

vsutil.py
OrangeChannel: Addressed in 4e2915a5f27080fbb1366c1607f8c8fc912058a5. Let me know if you have any suggestions for the two functions I added, I also added small test cases for them. I think _resolve_enum satisfies what kageru suggested. Keeping _readable_enums as its own function for now as it might be useful in the future separately from the _resolve_enum use.

vsutil.py
stuxcrystal: I think I will move this into a separate issue. No need to actually talk about here.

vsutil.py
```diff
@@ -176,3 +195,74 @@ def is_image(filename: str, /) -> bool:
...
+                             range_in=range_in,
+                             dither_type=dither_type.value)
+
+
+def _readable_enums(enum: EnumMeta, module: str = 'vsutil') -> list:
+    """
+    Returns a list of all possible values in `module.enum`.
+    Extends the default `repr(enum.value)` behavior by prefixing the enum with the name of the module it belongs to.
+    """
+    return [f'<{module}.{str(e)}: {e.value}>' for e in enum]
```
stuxcrystal, line 253: All classes are already recording the module of their creation in an attribute called `__module__`
Perhaps get module from `enum.__module__`? That way you don't need to pass the module name manually.

vsutil.py
```diff
@@ -176,3 +195,74 @@ def is_image(filename: str, /) -> bool:
...
+                     or (curr_depth > bitdepth and sample_type == vs.INTEGER))
+    dither_type = fallback(dither_type, Dither.ERROR_DIFFUSION if should_dither else Dither.NONE)
+
+    return clip.resize.Point(format=clip.format.replace(bits_per_sample=bitdepth, sample_type=sample_type),
+                             range=range,
+                             range_in=range_in,
+                             dither_type=dither_type.value)
+
+
+def _readable_enums(enum: EnumMeta, module: str = 'vsutil') -> list:
```
stuxcrystal, line 248: I think using `typing.Type[Enum]` is better readable for users as new script users might not know that all `Enum`-subclasses are actually instances of `EnumMeta`.

What about this signature:
`def _readable_enums(enum: Type[Enum]) -> List[str]:`

I removed the `module`-parameter. See a comment further down.

vsutil.py
```diff
@@ -176,3 +195,74 @@ def is_image(filename: str, /) -> bool:
...
+    """
+    Returns a list of all possible values in `module.enum`.
+    Extends the default `repr(enum.value)` behavior by prefixing the enum with the name of the module it belongs to.
+    """
+    return [f'<{module}.{str(e)}: {e.value}>' for e in enum]
+
+
+def _resolve_enum(enum: EnumMeta, value: Any, var_name: str, module: str = 'vsutil'):
```
stuxcrystal, line 256: See comment on the signature of `_readable_enums`.

vsutil.py
OrangeChannel: See: <https://p.kageru.moe/Ob8CF5.py>

Setting either `range` or `range_in` to `None` *or* just leaving them unset seems to have no difference in output frameprops so `core.resize` does seem to handle this correctly (or at least how I'd expect it to).

vsutil.py
OrangeChannel: Doesn't seem to be possible to do with imported modules nor local enums as this returns `__main__` as the module name. See: <https://p.kageru.moe/zRFhTw.py>

vsutil.py
OrangeChannel:
```suggestion
def _readable_enums(enum: EnumMeta, module: str = 'vsutil') -> List[str]:
```
I agree with this signature change but the other part is a bit weird and possible unnecessary since users of vsutil shouldn't be reading/using these private functions anyways.

vsutil.py
OrangeChannel:
```suggestion
def _resolve_enum(enum: EnumMeta, value: Any, var_name: str, module: str = 'vsutil') -> Any:
```
I think this is also relevant but not sure if there's a better way of handling this.

OrangeChannel: So I found a proper solution to the return type:
```py
E = TypeVar('E', bound=EnumMeta)
```
```suggestion
def _resolve_enum(enum: EnumMeta, value: Any, var_name: str, module: str = 'vsutil') -> Union[E, None]:
```

OrangeChannel: Discussed with Fichte on TeamSpeak, https://github.com/Irrational-Encoding-Wizardry/vsutil/pull/29/commits/616457b1472029aa8db63939bd0eb1b9c9e8de09 resolves these issues.

vsutil.py
OrangeChannel: See 5b00bd0fb675509af0bd95e20b6cff2f389e7868. Getting the `__module__` attr from a vs class returns 'importlib._bootstrap' (likely because of Cython) so I had to make a check for this and replace this with 'vapoursynth' as the module name.

vsutil.py
```diff
@@ -246,15 +246,16 @@ def depth(clip: vs.VideoNode,
                              dither_type=dither_type.value)
 
 
-def _readable_enums(enum: Type[Enum], module: str = 'vsutil') -> List[str]:
+def _readable_enums(enum: Type[Enum]) -> List[str]:
     """
     Returns a list of all possible values in `module.enum`.
     Extends the default `repr(enum.value)` behavior by prefixing the enum with the name of the module it belongs to.
     """
+    module = 'vapoursynth' if 'importlib' in enum.__module__ else enum.__module__
```
FichteFoll, line 254: I still think going with an optional module parameter to override is better, as it delegates the responsibility or removing the ambiguity (caused by the importlib._bootstrap value) to the caller, who has the required knowledge to properly resolve the knowledge gap. If an override is not provided, it should fall back to `enum.__module__`.

Really nitpicky, though.

vsutil.py
```diff
@@ -246,15 +246,16 @@ def depth(clip: vs.VideoNode,
                              dither_type=dither_type.value)
 
 
-def _readable_enums(enum: Type[Enum], module: str = 'vsutil') -> List[str]:
+def _readable_enums(enum: Type[Enum]) -> List[str]:
```
FichteFoll, line 249: Considering how the return value of this function is supposed to be used inside a string anyway and to be read by a human, why not return a string from this function in the first place? You'll never use the 'list' aspect of the return value.

FichteFoll:
Also, please don't squash this PR again, @kageru, because I believe the commit history to be valuable to indicate the progress it went through and has useful commit messages-

vsutil.py
OrangeChannel: Accepted feedback in 8473e5dd4b1f8c5d0bade1a08e5ea03e94777dee.

OrangeChannel:
Alright, all feedback addressed. All tests passed. Ready for merge.

```
============================= test session starts ==============================
platform linux -- Python 3.8.2, pytest-5.4.2, py-1.8.1, pluggy-0.13.1 -- /usr/bin/python3.8
cachedir: .pytest_cache
rootdir: /home/stalled/Documents/vsutil
collecting ... collected 14 items

test_vsutil.py::VsUtilTests::test_depth PASSED                           [  7%]
test_vsutil.py::VsUtilTests::test_fallback PASSED                        [ 14%]
test_vsutil.py::VsUtilTests::test_frame2clip PASSED                      [ 21%]
test_vsutil.py::VsUtilTests::test_get_depth PASSED                       [ 28%]
test_vsutil.py::VsUtilTests::test_get_w PASSED                           [ 35%]
test_vsutil.py::VsUtilTests::test_get_y PASSED                           [ 42%]
test_vsutil.py::VsUtilTests::test_insert_clip PASSED                     [ 50%]
test_vsutil.py::VsUtilTests::test_is_image PASSED                        [ 57%]
test_vsutil.py::VsUtilTests::test_iterate PASSED                         [ 64%]
test_vsutil.py::VsUtilTests::test_plane_size PASSED                      [ 71%]
test_vsutil.py::VsUtilTests::test_readable_enums PASSED                  [ 78%]
test_vsutil.py::VsUtilTests::test_resolve_enum PASSED                    [ 85%]
test_vsutil.py::VsUtilTests::test_split_join PASSED                      [ 92%]
test_vsutil.py::VsUtilTests::test_subsampling PASSED                     [100%]

============================== 14 passed in 0.11s ==============================
```

OrangeChannel:
Okay, *now* it's ready for merge. Had to be re-based ontop of the https://github.com/Irrational-Encoding-Wizardry/vsutil/commit/e5e702b486c13cb720cb751fcf128379c65a21f9 "Allow gray inputs for get_y " from kageru.

PR #33, opened 2020-05-29
Switch from CircleCI to GH-Actions

stuxcrystal:


kageru:
Is this ready for merge? I don’t know much about GH actions, but it looks fine to me.

Also, can I squash this? I’d rather not have 35 commits for CI setup in the history.

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,38 @@
...
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+        vs-version: [latest,50, 49]
```
kageru, line 18: Do we really need to check compatibility with 3 different versions? Latest is fine, but someone would have to update the `50, 49` whenever a new version is released. I don’t want to do that.

FichteFoll:
> Also, can I squash this?

Definitely, imo.

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,38 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Run tests
+
+on:
+  push:
+    branches: [ master ]
```
FichteFoll, line 8: Why only run this for the master branch? Same for pull requests against master.

stuxcrystal: Right now we don't have any other branch, do we?

.github/workflows/run-tests.yml
stuxcrystal: I'd rather make sure we have some policy that defines which versions each vsutil version supports. Putting it here cements this policy.

I'd like to make sure that a user doesn't have to switch to a new version immediately, as sometimes VapourSynth releases are broken. At the same time I want to have a codified place that authoritatively defines the supported versions.

stuxcrystal:
@kageru It's ready to merge. I added commits to switch to my new github-action that installs vapoursynth (44-50 right now) for you.

.github/workflows/run-tests.yml
FichteFoll: Currently, yes, but if you were to ever develop on a different branch, even for pull requests (on this repo), the action would not run until the PR was created.

PR #36, opened 2020-05-30
Convert vsutil to a package.

stuxcrystal:


kageru:
LGTM, but I don’t really feel qualified since this is about Python packaging. @FichteFoll do you have more experience with that?

OrangeChannel:
Should the version number be bumped to `0.3.0` since there's already a `0.2.0` in the releases tab?

stuxcrystal:
Probably, but I feel it is outside the scope of the PR.

setup.py
```diff
@@ -21,7 +21,10 @@ def run_tests(self):
 setup(
     name='vsutil',
     version='0.1.0',
-    py_modules=['vsutil'],
+    packages=find_packages(exclude=['tests']),
+    package_data={
+        'vsutil': ['py.typed']
+    },
     url='https://encode.moe/vsutil',
```
FichteFoll, line 28: This url should be fixed since it currently 404s.

heicrd:
I feel like it'd be cleaner to move `vsutil.py` to `vsutil/vsutil.py` and have `__init__.py` be
```python
"""
"""
from .vsutil import *
__all__ = ['Dither', 'Range', 'depth', 'fallback', 'frame2clip', 'get_depth', 'get_plane_size',
           'get_subsampling', 'get_w', 'get_y', 'insert_clip', 'is_image', 'iterate', 'join', 'plane', 'split']
```
rather than shoving the entire package into `__init__.py`

setup.py
stuxcrystal: Let me fix that.

stuxcrystal:
>  Or as an alternative to `__all__`, you can just import the specific attributes from .vsutil

`multiprocessing actually `import *` every module and concatenates the `__all__` of each submodule in the package inside `__init__.py`.

stuxcrystal:
I moved the work on that minimal-import solution to a different branch because stuff kept breaking especially the tests for resolve_enum.

I moved them to the branch stux-minimal-init.

setup.py
stuxcrystal: Fixed.

FichteFoll:
> I feel like it'd be cleaner to move `vsutil.py` to `vsutil/vsutil.py` and have `__init__.py` be […]
> rather than shoving the entire package into `__init__.py`

I personally see no need for this, as you still have the entire module/package contents in one file and just add one level of indirection. There is definitely a use case for splitting a package into sub-modules and exporting those on the `package.__init__` level, but this is not happening currently.
Doesn't really matter, either way.

OrangeChannel:
I think it looks a bit odd to a non-Python programmer to have the source code in a strangely-named file in a folder instead of just directly naming the file (or to a VapourSynth func/script writer that is used to just importing from a .py file what they need). But Fichte’s right that for a one-file package like this, leaving stuff in the `__init__` is just easier (and apparently won’t break the tests either).

heicrd:
I think that even as a python programmer it looks a bit odd, as `__init__.py` is intended to be reserved for package initialization.

stuxcrystal:
What about splitting this into two parts:
1. This PR. Makes it a package and declares it as typed.
2. Then we split the library up into smaller logical parts

stuxcrystal:
@kageru 
Since there is no discussion anymore, can you merge it?

PR #37, opened 2020-06-04
Add function/decorator checks for variable-format/resolution clips

OrangeChannel:
This prevents us having to write the same `ValueError` for every function that requires a clip to have a format because it tries to access the `clip.format`'s attributes.

### DIFF
--- a/vsutil/__init__.py
+++ b/vsutil/__init__.py
@@ -59,6 +59,7 @@ def is_variable(clip: vs.VideoNode, /, *, format: bool = False, resolution: bool
 def disallow_variable_format(function: Callable[..., R]) -> Callable[..., R]:
     """
     Function decorator that raises an exception if the input clip has a variable format.
+    Decorated function's first parameter must be of type `vapoursynth.VideoNode` and is the only parameter checked.
     """
     @wraps(function)
     def _check(clip: vs.VideoNode, *args, **kwargs) -> R:
@@ -71,6 +72,7 @@ def disallow_variable_format(function: Callable[..., R]) -> Callable[..., R]:
 def disallow_variable_resolution(function: Callable[..., R]) -> Callable[..., R]:
     """
     Function decorator that raises an exception if the input clip has a variable resolution.
+    Decorated function's first parameter must be of type `vapoursynth.VideoNode` and is the only parameter checked.
     """
     @wraps(function)
     def _check(clip: vs.VideoNode, *args, **kwargs) -> R:
--- a/vsutil/__init__.py
+++ b/vsutil/__init__.py
@@ -52,17 +52,17 @@ def disallow_variable_format(function: F) -> F:
     return cast(F, _check)
 
 
-def disallow_variable_resolution(function: Callable[..., R]) -> Callable[..., R]:
+def disallow_variable_resolution(function: F) -> F:
     """
     Function decorator that raises an exception if the input clip has a variable resolution.
     Decorated function's first parameter must be of type `vapoursynth.VideoNode` and is the only parameter checked.
     """
     @wraps(function)
-    def _check(clip: vs.VideoNode, *args, **kwargs) -> R:
+    def _check(clip: vs.VideoNode, *args, **kwargs) -> Any:
         if 0 in (clip.width, clip.height):
             raise ValueError('Variable-resolution clips not supported.')
         return function(clip, *args, **kwargs)
-    return _check
+    return cast(F, _check)
 
 
 @disallow_variable_format
--- a/vsutil/__init__.py
+++ b/vsutil/__init__.py
@@ -41,6 +41,7 @@ def get_subsampling(clip: vs.VideoNode, /) -> Union[None, str]:
     Returns the subsampling of a VideoNode in human-readable format.
     Returns None for formats without subsampling.
     """
+    _disallow_variable_format(clip)
     if clip.format.color_family not in (vs.YUV, vs.YCOCG):
         return None
     if clip.format.subsampling_w == 1 and clip.format.subsampling_h == 1:
@@ -63,6 +64,7 @@ def get_depth(clip: vs.VideoNode, /) -> int:
     """
     Returns the bit depth of a VideoNode as an integer.
     """
+    _disallow_variable_format(clip)
     return clip.format.bits_per_sample
 
 
@@ -133,6 +135,7 @@ def plane(clip: vs.VideoNode, planeno: int, /) -> vs.VideoNode:
     :param planeno:  The index that specifies which plane to extract.
     :return: A grayscale clip that only contains the given plane.
     """
+    _disallow_variable_format(clip)
     if clip.format.num_planes == 1 and planeno == 0:
         return clip
     return core.std.ShufflePlanes(clip, planeno, vs.GRAY)
@@ -153,6 +156,7 @@ def split(clip: vs.VideoNode, /) -> List[vs.VideoNode]:
     """
     Returns a list of planes for the given input clip.
     """
+    _disallow_variable_format(clip)
     return [plane(clip, x) for x in range(clip.format.num_planes)]
 
 
@@ -232,6 +236,7 @@ def depth(clip: vs.VideoNode,
 
     :return: Converted clip with desired bit depth and sample type. ColorFamily will be same as input.
     """
+    _disallow_variable_format(clip)
     sample_type = _resolve_enum(vs.SampleType, sample_type, 'sample_type', 'vapoursynth')
     range = _resolve_enum(Range, range, 'range')
     range_in = _resolve_enum(Range, range_in, 'range_in')
--- a/vsutil/__init__.py
+++ b/vsutil/__init__.py
@@ -1,14 +1,15 @@
 """
 VSUtil. A collection of general-purpose VapourSynth functions to be reused in modules and scripts.
 """
-__all__ = ['Dither', 'Range', 'depth', 'fallback', 'frame2clip', 'get_depth', 'get_plane_size',
+__all__ = ['Dither', 'Range', 'depth', 'disallow_variable_format', 'disallow_variable_resolution', 'fallback',
+           'frame2clip', 'get_depth', 'get_plane_size',
            'get_subsampling', 'get_w', 'get_y', 'insert_clip', 'is_image', 'is_variable', 'iterate', 'join', 'plane',
            'split']
 
 from enum import Enum, IntEnum
 from mimetypes import types_map
 from os import path
-from typing import Any, Callable, List, Literal, Optional, Tuple, Type, TypeVar, Union
+from typing import Any, Callable, List, Literal, Optional, Tuple, Type, TypeVar, Union, cast
 
 import vapoursynth as vs
 core = vs.core
@@ -16,6 +17,7 @@ core = vs.core
 T = TypeVar('T')
 E = TypeVar('E', bound=Enum)
 R = TypeVar('R')
+C = TypeVar('C', bound=Callable)
 
 
 class Range(IntEnum):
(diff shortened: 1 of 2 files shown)


vsutil/__init__.py
```diff
@@ -275,3 +280,13 @@ def _resolve_enum(enum: Type[E], value: Any, var_name: str, module: Optional[str
         return enum(value)
     except ValueError:
         raise ValueError(f'{var_name} must be in {_readable_enums(enum, module)}.') from None
+
+
+def _disallow_variable(clip: vs.VideoNode, /) -> None:
```
stuxcrystal, line 285: What about this signature:
```py
def _disallow_variable(clip: vs.VideoNode, *, format: bool = True, resolution=True) -> None
```

It would be more reusable.

vsutil/__init__.py
```diff
@@ -2,7 +2,8 @@
 VSUtil. A collection of general-purpose VapourSynth functions to be reused in modules and scripts.
 """
 __all__ = ['Dither', 'Range', 'depth', 'fallback', 'frame2clip', 'get_depth', 'get_plane_size',
-           'get_subsampling', 'get_w', 'get_y', 'insert_clip', 'is_image', 'iterate', 'join', 'plane', 'split']
+           'get_subsampling', 'get_w', 'get_y', 'insert_clip', 'is_image', 'is_variable', 'iterate', 'join', 'plane',
+           'split']
```
stuxcrystal, line 6: Off-Topic, but:

This is ruby-code-style and this is horrible. Python demands:

```py
my_list = [
    "a",
    "b",
    ...
]

Comment (OrangeChannel) [reply vsutil/__init__.py:6]: I just auto-formatted the list with PyCharm, we can re-format this in another PR.
```py
__all__ = [
    'Dither', 'Range',

    'depth',
    'fallback',
    'frame2clip',
    'insert_clip',
    'iterate',
    'join',
    'plane',
    'split',

    'get_depth', 'get_plane_size', 'get_subsampling', 'get_w', 'get_y',

    'is_image', 'is_variable'
]
```
Would this be acceptable? Tt groups a few of them logically and looks better, if not we can one-per-line the whole list.

FichteFoll: I'm good with grouping personally, but the larger section in the middle could also be grouped in some ways.
Just remember to add a trailing comma.

vsutil/__init__.py
```diff
@@ -275,3 +281,31 @@ def _resolve_enum(enum: Type[E], value: Any, var_name: str, module: Optional[str
...
+
+
+def _disallow_variable_format(clip: vs.VideoNode, /) -> None:
+    if is_variable(clip, format=True):
+        raise ValueError('Variable-format clips not supported.')
+
+
+def _disallow_variable_resolution(clip: vs.VideoNode, /) -> None:
+    if is_variable(clip, resolution=True):
+        raise ValueError('Variable-resolution clips not supported.')
```
FichteFoll, line 311: I suggest turning these into decorators. Makes the intent clear and currently they are all used as the first line of the function anyway.

OrangeChannel: ooooh, good idea :+1: , will implement later.

OrangeChannel: Implemented in https://github.com/Irrational-Encoding-Wizardry/vsutil/pull/37/commits/03f90fce8bd4d8235dcf913c4d49a52a3a00699e, not sure if this is the correct way to do this since I've never created a decorator before, nevermind a properly type hinted one. Maybe @heicrd can weigh in on the type hinting. mypy seems to be happy with this, using `typing.Callable` was a mess as the function signature is not guaranteed.

OrangeChannel: Oh, nope this is actually broken, welp.

OrangeChannel: I'm dumb, here, fixed (I think) https://github.com/Irrational-Encoding-Wizardry/vsutil/pull/37/commits/95beb7fad0bd680597248ddbaa67b97a05fcba21

I still need to add tests for this but I've tested it locally and it works as expected, just not sure this is the correct way of doing this.

vsutil/__init__.py
stuxcrystal: I'm fine with grouping too.

But formatting lists like this:
```py
my_list = ["123", "456", ...,
                "abc", "def"]
```

Is in my experience far too unstable, that's why I hate that syntax.

vsutil/__init__.py
```diff
@@ -35,6 +37,47 @@ class Dither(Enum):
...
+            return True
+
+    return False
+
+
+def disallow_variable_format(function: Callable[..., R]) -> Callable[..., R]:
+    """
+    Function decorator that raises an exception if the input clip has a variable format.
+    """
+    def _check(clip: vs.VideoNode, *args, **kwargs) -> R:
```
stuxcrystal, line 62: Add

```py
@functools.wraps(function)
```

to the wrapper function (`_check`).

vsutil/__init__.py
```diff
@@ -1,13 +1,15 @@
...
+__all__ = ['Dither', 'Range', 'depth', 'disallow_variable_format', 'disallow_variable_resolution', 'fallback',
+           'frame2clip', 'get_depth', 'get_plane_size',
+           'get_subsampling', 'get_w', 'get_y', 'insert_clip', 'is_image', 'is_variable', 'iterate', 'join', 'plane',
+           'split']
 
 from enum import Enum, IntEnum
 from mimetypes import types_map
 from os import path
-from typing import Any, Callable, List, Literal, Optional, Tuple, Type, TypeVar, Union
+from typing import Any, Callable, List, Literal, Optional, Tuple, Type, TypeVar, Union, cast
```
stuxcrystal, line 12: Why do you need `cast`

vsutil/__init__.py
```diff
@@ -35,6 +37,47 @@ class Dither(Enum):
...
+            raise ValueError('Variable-format clips not supported.')
+        return function(clip, *args, **kwargs)
+    return _check
+
+
+def disallow_variable_resolution(function: Callable[..., R]) -> Callable[..., R]:
+    """
+    Function decorator that raises an exception if the input clip has a variable format.
+    """
+    def _check(clip: vs.VideoNode, *args, **kwargs) -> R:
```
stuxcrystal, line 73: Add `functools.wraps`

heicrd:
Is there a way to get mypy to recognize this? Right now it doesn't seem smart enough to follow the `None` check through the decorator.

stuxcrystal:
@heicrd yes. `typing.cast`

vsutil/__init__.py
OrangeChannel: Forgot to remove from the previous commit.

vsutil/__init__.py
OrangeChannel: Added in https://github.com/Irrational-Encoding-Wizardry/vsutil/pull/37/commits/f942ce28bc67a88992a87409adc07f55c036c99a, is this so `function.__name__` and `function.__doc__` still work?

OrangeChannel:
This can all be squashed into one commit with the PR title and reference \#.

OrangeChannel:
Should we bump the version number in this update? This technically changes the exposed API by adding three new functions.
I don't know how git-tagging works for PRs or forks, I'd assume pushing tags directly to upstream/master is the better or only way to do it.

I recommend tagging ee49d9d16e70fc3615a84ea0da02748f3ae71ba3 as `0.3.0`, 
(and af4dd0a35f2cfb6414f340e04cf40bb5dcdaa249 as `0.3.1` is we want to be specific), and whatever this squashed commit becomes as `0.4.0`.

vsutil/__init__.py
```diff
@@ -35,6 +38,51 @@ class Dither(Enum):
     ERROR_DIFFUSION = 'error_diffusion'  # Floyd-Steinberg error diffusion.
 
 
+def is_variable(clip: vs.VideoNode, /, *, format: bool = False, resolution: bool = False) -> bool:
```
heicrd, line 41: Not only do i feel like this logic isn't complex enough to warrant its own function, having to specify which you want to check for (as apposed to having two separate functions and making the functions even more redundant) makes it extra unwieldy.

OrangeChannel: Agreed, if another script author wants to do something other than use the decorators with this logic it'd be easier for them to just write out the two lines `if clip.format is None; if 0 in (clip.width, clip.height)`.

vsutil/__init__.py
```diff
@@ -36,12 +38,38 @@ class Dither(Enum):
     ERROR_DIFFUSION = 'error_diffusion'  # Floyd-Steinberg error diffusion.
 
 
+def disallow_variable_format(function: Callable[..., R]) -> Callable[..., R]:
```
FichteFoll, line 41: For proper type preservation, you should use a typevar for `Callable[..., R]`. See https://mypy.readthedocs.io/en/stable/generics.html#declaring-decorators.

OrangeChannel: https://github.com/Irrational-Encoding-Wizardry/vsutil/pull/37/commits/93324264cd82cb6afaeee010657747b6f2e64862 look good?

OrangeChannel:
So, I've completely satisfied mypy's issues on vsutil's side here: https://www.diffchecker.com/tGKtXhE5
Not sure if this fits in this PR, or is even wanted but figured I'd show @stuxcrystal @heicrd.

kageru:
> So, I've completely satisfied mypy's issues on vsutil's side here: https://www.diffchecker.com/tGKtXhE5
> Not sure if this fits in this PR, or is even wanted but figured I'd show @stuxcrystal @heicrd.

I’m against doing that. Having the casts inside the decorators is fine, but adding them everywhere just to make a dysfunctional typechecker happy is not what I want. We shouldn’t make code more complicated just to please the IDE or some third-party tool.
Having decorators at all is already a level of complexity that might scare away potential contributors (most VS users don’t know much about Python), but I think they’re quite self-explanatory in this case, so I’m fine with that.
  

In general, we might consider making some of these functions work with variable formats (e.g. I don’t see a reason why `depth` shouldn’t work on a variable-depth input), but that’s out of scope here.

As for the tags, I’ve tagged the commit you suggested as 0.3.0, and I agree with tagging this as 0.4.0. I’ll do that after merge.
The decorators are part of the public API, and I believe they can be useful to other script and library authors, so I’d definitely consider that a new feature, and thus a minor version bump.
We technically also change the existing API, but in a very minor way (i.e. always raising `ValueError`s, where previously an unintended `AttributeError` could have occured), but I don’t think that matters in practive due to how Vapoursynth scripts are usually written.

heicrd:
Yeah you can just cast inside the decorator and change the function signature to `def fun(clip: _ConstantFormatVideoNode, ...)` etc without having to use casts anywhere.

>Having decorators at all is already a level of complexity that might scare away potential contributors

vsutil is already beyond this point with using typehints and unit tests in the first place imo. i don't think it's a good idea either way to try and draw contributors who have actually zero programming experience

vsutil/__init__.py
```diff
@@ -1,20 +1,24 @@
...
-from typing import Any, Callable, List, Literal, Optional, Tuple, Type, TypeVar, Union
+from typing import Any, Callable, cast, List, Literal, Optional, Tuple, Type, TypeVar, Union
 
 import vapoursynth as vs
 core = vs.core
 
 T = TypeVar('T')
 E = TypeVar('E', bound=Enum)
 R = TypeVar('R')
+F = TypeVar('F', bound=Callable[..., Any])
```
stuxcrystal, line 21: Can you try to mark this as covariant.

OrangeChannel:
I’ll make a more coherent response later but wanted to say that casting inside the decorator doesn’t solve mypy’s problem. It also creates another problem where another script calling get_subsampling(clip_VideoNode) is an error because isinstance(clip_VideoNode, _ConstantFormatNode) is False. mypy isn’t smart enough to know what to do with a changed signature via the decorator and casting inside the decorator means losing the ability to use the decorated functions with normal VideoNodes.

vsutil/__init__.py
OrangeChannel:
```
__init__.py:42: error: Cannot use a covariant type variable as a parameter
__init__.py:55: error: Cannot use a covariant type variable as a parameter
```
This is what I get after doing that, whats the reason behind this change anyways?

stuxcrystal: covariant marks a parameter as "accepts subtypes too".

OrangeChannel: image
Doesn't `bound` in this case accomplish what we want?

stuxcrystal: I see. I think it was worth a try though.

kageru:
> I’ll make a more coherent response later

I’m still kind of waiting for that.
Are there any other changes you still want to make here or can we merge this?

OrangeChannel:
Oh I think I just explained the problem with typechecking the decorators in Light’s discord. There’s not much we can do here that wouldn’t be super over-engineered so as far as I’m concerned, the PR is ready to merge.

OrangeChannel:
There’s a few open issues in mypy’s repo that might address this stuff in the future but for now the casting and stuff isn’t worth it.

PR #41, opened 2020-06-17
Cleanup dithering conditional and explanation

OrangeChannel:
Shouldn't change anything because calling resize with float output and a `dither_type` disregarded the `dither_type` value anyways. Reverts some of the changed behavior in #39 and avoids the misleading `should_dither` being true for float sample type.

### DIFF
--- a/vsutil/__init__.py
+++ b/vsutil/__init__.py
@@ -303,7 +303,7 @@ def _should_dither(out_bits: int,
     """
     out_sample_type = fallback(out_sample_type, vs.FLOAT if out_bits == 32 else vs.INTEGER)
 
-    if out_sample_type == vs.FLOAT:  # zimg will not dither for conversions resulting in float sample type
+    if out_sample_type == vs.FLOAT:
         return False
 
     if clip is not None:
--- a/vsutil/__init__.py
+++ b/vsutil/__init__.py
@@ -7,7 +7,7 @@ __all__ = ['Dither', 'Range', 'depth', 'fallback', 'frame2clip', 'get_depth', 'g
 from enum import Enum, IntEnum
 from mimetypes import types_map
 from os import path
-from typing import Any, Callable, List, Literal, Optional, Tuple, Type, TypeVar, Union
+from typing import Any, Callable, cast, List, Literal, Optional, Tuple, Type, TypeVar, Union
 
 import vapoursynth as vs
 core = vs.core
@@ -227,7 +227,7 @@ def depth(clip: vs.VideoNode,
     :param range_in:    Input pixel range (defaults to input clip's range). See `Range`.
     :param dither_type: Dithering algorithm. Allows overriding default dithering behavior. See `Dither`.
         Defaults to Floyd-Steinberg error diffusion when downsampling, converting between ranges, or upsampling full
-        range input. Defaults to 'none', or round to nearest, otherwise.
+        range input. Defaults to 'none', or round to nearest, otherwise. See `_should_dither()` for more information.
 
     :return: Converted clip with desired bit depth and sample type. ColorFamily will be same as input.
     """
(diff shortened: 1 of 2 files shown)


vsutil/__init__.py
```diff
@@ -244,13 +244,16 @@ def depth(clip: vs.VideoNode,
...
-                     or curr_depth > bitdepth)
+    # The exception is when converting from 8 to 16 bit full range, (0-255) * 257 -> (0-65535).
+    # Dithering is always needed when converting from float to integer precision.
+    # Dithering is theoretically needed when converting from an integer depth greater than 10 to half float,
+    # despite the higher bitdepth, but zimg currently will not dither for float output.
+    should_dither = (range_in != range  # full->limited OR limited->full
+                     or clip.format.sample_type == vs.FLOAT  # float->int
+                     or (range_in == Range.FULL  # full/int->full/int...
+                         and not (curr_depth, bitdepth) == (8, 16))  # ...but specifically not for 8->16
+                     or curr_depth > bitdepth) and sample_type == vs.INTEGER  # limited/int->limited/int if downsampling
```
FichteFoll, line 255: `sample_type == vs.INTEGER` should be the first condition, because it is required and simple and should most definitely *not* be on the same line as the end of the large disjunction.

OrangeChannel:
```py
...
    should_dither = (sample_type == vs.INTEGER
                     and (range_in != range  # full->limited OR limited->full
                          or clip.format.sample_type == vs.FLOAT  # float->int
                          or (range_in == Range.FULL  # full/int->full/int...
                              and (curr_depth, bitdepth) != (8, 16))  # ...but specifically not for 8->16
                          or curr_depth > bitdepth))  # limited/int->limited/int if downsampling
```
This look good?

FichteFoll: Yes, this is better. I would argue that most of the comments seem rather self-explanatory and don't add to understandability of the code. In fact, is ` limited/int->limited/int if downsampling` even correct, because that condition can also be reached when `(range_in, range) == (Range.FULL, Range.FULL)`?

stuxcrystal:
We are at our third PR with discussion about this topic iirc.

At this point I think we should split the one-liner into multiple lines and put them into a (possibly internal) function.

vsutil/__init__.py
OrangeChannel: I refactored this into its own function with tests now.

OrangeChannel:
This can get squashed into something like `fix dithering logic (#41)` and tagged as `0.3.1`. I'll then rebase #37 ontop of this since it shares some of the test code and depth needs the decorator too.

Or merge #37 first (tagged as `0.4.0`) and then I'll rebase this and get this tagged as `0.4.1` (which should be the next pypi/package release).

vsutil/__init__.py
```diff
@@ -282,3 +272,57 @@ def _resolve_enum(enum: Type[E], value: Any, var_name: str, module: Optional[str
...
+    range_conversion = in_range != out_range
+    float_to_int = in_sample_type == vs.FLOAT
+    upsampling = cast(int, in_bits) < out_bits
+    downsampling = cast(int, in_bits) > out_bits
+
+    return True if (range_conversion
+                    or float_to_int
+                    or (in_range == Range.FULL and upsampling and (in_bits, out_bits) != (8, 16))
+                    or downsampling) \
+        else False
```
FichteFoll, line 328:
```suggestion
    return bool(range_conversion
                or float_to_int
                or (in_range == Range.FULL and upsampling and (in_bits, out_bits) != (8, 16))
                or downsampling)
```

vsutil/__init__.py
```diff
@@ -279,3 +272,56 @@ def _resolve_enum(enum: Type[E], value: Any, var_name: str, module: Optional[str
         return enum(value)
     except ValueError:
         raise ValueError(f'{var_name} must be in {_readable_enums(enum, module)}.') from None
+
+
+def _should_dither(out_bits: int,
```
kageru, line 277: You converted a 4-line condition into a 50 line helper function. This is way overkill.

vsutil/__init__.py
```diff
@@ -279,3 +272,56 @@ def _resolve_enum(enum: Type[E], value: Any, var_name: str, module: Optional[str
...
+                   clip: Optional[vs.VideoNode] = None,
+                   in_bits: Optional[int] = None,
+                   in_range: Optional[Range] = None,
+                   in_sample_type: Optional[vs.SampleType] = None,
+                   ) -> bool:
+    """
+    Determines whether dithering is needed for a given depth/range/sample_type conversion.
+
+    If a clip is given, the clip's properties will take precedence over the in_bits and in_sample_type parameters.
+    If a variable-format clip is given, in_bits and in_sample_type are required.
```
kageru, line 289: This is an internal helper function that is only used once and will likely never be used elsewhere.
It’s not necessary to have multiple ways of passing parameters. It’s just needlessly complicated.

vsutil/__init__.py
OrangeChannel: Now 15 loc, most of this is the params not fitting on one line and the explanation that was originally in-line comments.

PR #42, opened 2020-06-29
Various fixes

OrangeChannel:
Addresses https://github.com/Irrational-Encoding-Wizardry/vsutil/pull/37#discussion_r435503076 and a couple other issues found with mypy and whatever PyCharm uses.

### DIFF
--- a/vsutil/__init__.py
+++ b/vsutil/__init__.py
@@ -1,10 +1,18 @@
 """
 VSUtil. A collection of general-purpose VapourSynth functions to be reused in modules and scripts.
 """
-__all__ = ['Dither', 'Range', 'depth', 'disallow_variable_format', 'disallow_variable_resolution', 'fallback',
-           'frame2clip', 'get_depth', 'get_plane_size',
-           'get_subsampling', 'get_w', 'get_y', 'insert_clip', 'is_image', 'iterate', 'join', 'plane',
-           'split']
+__all__ = [
+    # enums
+    'Dither', 'Range',
+    # decorators
+    'disallow_variable_format', 'disallow_variable_resolution',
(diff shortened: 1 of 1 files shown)


vsutil/__init__.py
```diff
@@ -279,13 +288,11 @@ def depth(clip: vs.VideoNode,
         return clip
 
     should_dither = _should_dither(curr_depth, bitdepth, range_in, range, clip.format.sample_type, sample_type)
-
     dither_type = fallback(dither_type, Dither.ERROR_DIFFUSION if should_dither else Dither.NONE)
```
stuxcrystal, line 292: Do we need so many blank lines?

OrangeChannel: This commit actually gets rid of that blank line and groups together the dither logic. :thinking:

vsutil/__init__.py
```diff
@@ -279,13 +288,11 @@ def depth(clip: vs.VideoNode,
...
 
-    return clip.resize.Point(format=clip.format.replace(bits_per_sample=bitdepth, sample_type=sample_type),
-                             range=range,
-                             range_in=range_in,
-                             dither_type=dither_type.value)
+    new_format = clip.format.replace(bits_per_sample=bitdepth, sample_type=sample_type).id
```
kageru, line 293: You call `.id` on the new format here which didn’t happen before. Are both correct, and if so, why change it now?

stuxcrystal: Because the typestubs do not include `Format` as a valid type to be passed to a vapoursynth filter.
It is supported though.
Mypy does however does not like it.

kageru: I see. Thank you.

PR #43, opened 2020-07-07
Use a special value to denote "No Value" in Fallback

stuxcrystal:


kageru:
What is actually the use case for this, i.e. when do you want `None` as the return value of `fallback`?

stuxcrystal:
The idea is that you can write something like this:

```py
def my_function(parameter: Optional[int] = no_value()):
    parameter = fallback(parameter, 10, none_as_default=False)
    if parameter is None: ...
```

I admit this is a rare case. I think this is most useful for parameters directly passed to plugins. If a function thinks it has a better default but allows to pass None which in vapoursynth-plugin denotes default value.

kageru:
If the function thinks it has a better default, wouldn’t it just do
```py
def my_function(parameter: int = 10):
    SomePlugin(parameter)
```
which would take the default value if the parameter is `None`. I don’t think fallback should be used here at all. It’s only for more complex fallback values, i.e. making `parameter2` equal to `parameter1` or generally having your default values depend on other parameters, in which case using the plugin default is probably not what the caller wants.

stuxcrystal:
> It’s only for more complex fallback values, i.e. making parameter2 equal to parameter1 or generally having your default values depend on other parameters, in which case using the plugin default is probably not what the caller wants.

Exactly. My PR is something more for cases like 

```py
_empty = object()

def func(y: VideoNode, iterations: Any=_empty):
    if x is _empty:  # default argument value
        x = 10
    if x is not None:  # argument was provided and it's None
        y = iterate(y, lambda y: y.std.Minimum().std.Maximum(), x)

    return y.some_plugin.do_stuff()
```

Fallback could now implement it like this:
```py
def func(y: VideoNode, x: Optional[int]=no_value()):
    x = fallback(x, my_default_for(y), none_as_default=False)
    if x is not None:
        y = iterate(y, lambda y: y.std.Minimum().std.Maximum(), x)
    return y.some_plugin.do_stuff()
```

I admit that `iterate` is a bad example as you could theoretically just set x to 0. But this example was supposed to show that you can use this to turn off features of a function by passing `None` to a function.

My solution makes it easier as `no_value()` is actually generic. Any necessity to use anEnum-Class like https://www.python.org/dev/peps/pep-0484/#id30 is removed that way and the type-annotation is more concise.

kageru:
Sorry, but I still don’t consider this generally useful. I don’t see a single use for it in vsutil, and having a parameter that could be `Some(x)` for a value, `empty()` for a default, or `None` to disable an optional feature altogether seems like a very confusing thing to do.

I’d even argue that needing this in a function might be a good indicator that the code should be refactored.

PR #44, opened 2020-07-08
Add "scale" function

kgrabs:
Made this helper function to scale pixel values in various ways to be used by plugins such as std.Expr and std.Binarize.


vsutil/__init__.py
```diff
@@ -251,6 +251,65 @@ def is_image(filename: str, /) -> bool:
     return types_map.get(path.splitext(filename)[-1], '').startswith('image/')
 
 
+def scale(value: float, 
+          input_depth: int = 8, 
+          output_depth: int = 32,
```
stuxcrystal, line 256: I feel like these are essential parameters that should always be passed.

vsutil/__init__.py
```diff
@@ -251,6 +251,65 @@ def is_image(filename: str, /) -> bool:
     return types_map.get(path.splitext(filename)[-1], '').startswith('image/')
 
 
+def scale(value: float, 
+          input_depth: int = 8, 
+          output_depth: int = 32, 
+          range_in: Union[int, Range] = False,
```
stuxcrystal, line 257: Why use booleans here and not ints? I think this is needlessly confusing.
I complain because `True` and `False` are ints, yes but they have additional connotations.

vsutil/__init__.py
```diff
@@ -251,6 +251,65 @@ def is_image(filename: str, /) -> bool:
...
+    :chroma:        Whether to treat values as chroma instead of luma
+    """
+
+    range_in = _resolve_enum(Range, range_in, 'range_in')
+    range = _resolve_enum(Range, range, 'range')
+    range = fallback(range, range_in)
+
+    if input_depth == 32:
+        input_peak = 1
+        range_in = True
```
stuxcrystal, line 279: See comment on range_in parameter.

vsutil/__init__.py
```diff
@@ -251,6 +251,65 @@ def is_image(filename: str, /) -> bool:
...
+    elif range_in:
+        input_peak = (1 << input_depth) - 1
+    else:
+        input_peak = (224 if chroma else 219) << (input_depth - 8)
+
+    if output_depth == 32:
+        output_peak = 1
+        range = True
```
stuxcrystal, line 287: See comment on range_in parameter.

vsutil/__init__.py
```diff
@@ -251,6 +251,65 @@ def is_image(filename: str, /) -> bool:
     return types_map.get(path.splitext(filename)[-1], '').startswith('image/')
 
 
+def scale(value: float,
```
stuxcrystal, line 254: This code doesn't even have dependencies on vapoursynth. Why not add a test?

vsutil/__init__.py
```diff
@@ -251,6 +251,65 @@ def is_image(filename: str, /) -> bool:
...
+    range = _resolve_enum(Range, range, 'range')
+    range = fallback(range, range_in)
+
+    if input_depth == 32:
+        input_peak = 1
+        range_in = 1
+    elif range_in:
+        input_peak = (1 << input_depth) - 1
+    else:
+        input_peak = (224 if chroma else 219) << (input_depth - 8)
```
kageru, line 283: If my memory (and also Wikipedia) serves me right, the peak values for limited range are 235 for luma and 240 for chroma. Am I missing something here?

vsutil/__init__.py
```diff
@@ -251,6 +251,65 @@ def is_image(filename: str, /) -> bool:
...
+    :chroma:        Whether to treat values as chroma instead of luma
+    """
+
+    range_in = _resolve_enum(Range, range_in, 'range_in')
+    range = _resolve_enum(Range, range, 'range')
+    range = fallback(range, range_in)
+
+    if input_depth == 32:
+        input_peak = 1
+        range_in = 1
```
kageru, line 279: I would prefer to set `range_in` in a separate block. Doing that means this if/elif chain would have no side effects other than setting `input_peak` which would allow us to extract the peak value logic into a small function to avoid the duplication.

vsutil/__init__.py
kageru, line 254: Will typecheckers complain about this not being Union[float, int]`?

kgrabs: I made the test, and I think it's thorough, but I have no idea how to run it.

vsutil/__init__.py
kgrabs: Those are the peaks with no YUV offsets applied, so 235 - 16 and 240 - 16

vsutil/__init__.py
kgrabs: Alright, everything looks good on that end, unless I missed a parameter combination somehow.

kgrabs: I guess I might as well change it just to be safe

kgrabs:
Is it possible to re-run the tests? The current version is renamed properly but the error report contains the old name

https://github.com/Irrational-Encoding-Wizardry/vsutil/pull/44/commits/0e3cde840334d51bb1d38a17b1e7772864e45c7c
```
 Traceback (most recent call last):
  File "D:\a\vsutil\vsutil\tests\test_vsutil.py", line 178, in test_scale
    self.assertEqual(vsutil.scale(1, 8, 8, range_in=0, range=0), 1)
AttributeError: module 'vsutil' has no attribute 'scale'
```

stuxcrystal:
Relevant section from the github action-config:
```yaml
    - name: Running tests
      run: |
        python -m unittest discover -s ./tests
```

kageru:
Looks good to me. Opinions on tagging this as 0.5.0?

stuxcrystal:
I'd wait until we split up the file into multiple submodules.

PR #46, opened 2020-07-18
Add Sphinx documentation

OrangeChannel:
I tried setting up a github action to run `make html` but I can't get it to work how I thought it would.

We'd have to build the docs to a gh-pages branch I think but I don't know what I'm doing.

<https://github.com/ammaraskar/sphinx-action#great-actions-to-pair-with>

This needs to be split into multiple commits probably, and *a lot* of the documentation is *way* too obvious and unneeded but I put it all anyways. Live version at <https://orangechannel.github.io/vsutil/html/index.html>.

### DIFF
--- a/docs/_static/css/theme_overrides.css
+++ b/docs/_static/css/theme_overrides.css
@@ -6,10 +6,6 @@
     }
 }
 
-.rst-content code,.rst-content tt {
-    font-family: SFMono-Regular,Menlo,Monaco,Consolas,"Liberation Mono","Courier New",monospace
-}
-
 .rst-content code.literal {
     color: #595959
 }
--- a/docs/_static/css/theme_overrides.css
+++ b/docs/_static/css/theme_overrides.css
@@ -1,3 +1,5 @@
+@import 'theme.css';
+
 @media screen and (min-width:1100px) {
     .wy-nav-content {
         max-width: 1000px
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1,0 +1,58 @@
+=============
+API Reference
+=============
+
+.. automodule:: vsutil
+
+
+Functions that return a clip
+============================
+
+.. autofunction:: vsutil.depth
+.. autofunction:: vsutil.frame2clip
+.. autofunction:: vsutil.get_y
+.. autofunction:: vsutil.insert_clip
+.. autofunction:: vsutil.join
+.. autofunction:: vsutil.plane
+.. autofunction:: vsutil.split
+
+
+Miscellanious non-VapourSynth functions
+=======================================
+
+.. autofunction:: vsutil.fallback
+.. autofunction:: vsutil.iterate
+
+
+Decorators
+==========
+
+.. autofunction:: vsutil.disallow_variable_format
+.. autofunction:: vsutil.disallow_variable_resolution
+
+
+Clip information and helper functions
+=====================================
+
+**Helpers to inspect a clip/frame**
+
+.. autofunction:: vsutil.get_depth
+.. autofunction:: vsutil.get_plane_size
+.. autofunction:: vsutil.get_subsampling
+.. autofunction:: vsutil.is_image
+
+----
+
+**Mathematical helpers**
+
+.. autofunction:: vsutil.get_w
+.. autofunction:: vsutil.scale_value
+
+
+Enums
+=====
+
+.. autoclass:: vsutil.Dither
+    :members:
+.. autoclass:: vsutil.Range
+    :members:
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,0 +1,75 @@
+# Configuration file for the Sphinx documentation builder.
+#
+# This file only contains a selection of the most common options. For a full
+# list see the documentation:
+# https://www.sphinx-doc.org/en/master/usage/configuration.html
+
+# -- Path setup --------------------------------------------------------------
+
+# If extensions (or modules to document with autodoc) are in another directory,
+# add these directories to sys.path here. If the directory is relative to the
+# documentation root, use os.path.abspath to make it absolute, like shown here.
+
+import os
+import sys
+
+sys.path.insert(0, os.path.abspath('..'))
+
+
+# -- Project information -----------------------------------------------------
+
+project = 'vsutil'
+copyright = '2020, Irrational Encoding Wizardry'
+author = 'Irrational Encoding Wizardry'
+
+# The full version, including alpha/beta/rc tags
+version = release = '0.5.0'
+
+
+# -- General configuration ---------------------------------------------------
+
+extensions = [
+    'sphinx.ext.autodoc',
+    'sphinx.ext.autosummary',
+    'sphinx.ext.todo',
+    'sphinx.ext.viewcode',
+    'sphinx_autodoc_typehints',
+]
+
+# Add any paths that contain templates here, relative to this directory.
+templates_path = ['_templates']
+
+# List of patterns, relative to source directory, that match files and
+# directories to ignore when looking for source files.
+# This pattern also affects html_static_path and html_extra_path.
+exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store']
+
+
+# -- Options for HTML output -------------------------------------------------
+
+# The theme to use for HTML and HTML Help pages.  See the documentation for
+# a list of builtin themes.
+#
+html_theme = 'sphinx_rtd_theme'
+
+# Add any paths that contain custom static files (such as style sheets) here,
+# relative to this directory. They are copied after the builtin static files,
+# so a file named "default.css" will overwrite the builtin "default.css".
+html_static_path = ['_static']
+
+html_context = {
+    # https://rackerlabs.github.io/docs-rackspace/tools/rtd-tables.html
+    'css_files': ['theme_overrides.css'],
+}
+
+autosummary_generate = True
+autodoc_mock_imports = ['vapoursynth']
+smartquotes = True
+html_show_sphinx = False
+# add_module_names = False
+pygments_style = 'sphinx'
+
+
+# -- Extension configuration -------------------------------------------------
+
+todo_include_todos = True
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -57,10 +57,11 @@ html_theme = 'sphinx_rtd_theme'
 # so a file named "default.css" will overwrite the builtin "default.css".
 html_static_path = ['_static']
 
-html_context = {
-    # https://rackerlabs.github.io/docs-rackspace/tools/rtd-tables.html
-    'css_files': ['theme_overrides.css'],
-}
+html_css_files = [
+    'css/theme_overrides.css'
+]
+
+html_style = 'css/theme_overrides.css'
 
 autosummary_generate = True
 autodoc_mock_imports = ['vapoursynth']
--- a/vsutil/clips.py
+++ b/vsutil/clips.py
@@ -1,5 +1,5 @@
 """
-_
+Functions that modify/return a clip.
 """
 __all__ = ['depth', 'frame2clip', 'get_y', 'insert_clip', 'join', 'plane', 'split']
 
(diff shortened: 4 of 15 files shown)


stuxcrystal:
Do we need to have the rendered documentation checked in?

OrangeChannel:
> Do we need to have the rendered documentation checked in?

Are you asking if we have to test if the documentation builds with no errors? If so, probably would be a good idea to.

stuxcrystal:
I am asking if we need to have .html files with the rendered documentation checked into the repository? Can we just keep the md and all other files neccessary for building the documentation and let the CI just push the result to a different branch (since GH-pages can deloy from a different branch)

OrangeChannel:
That would be ideal yea, everything in `docs/html/` can be built from the files in `docs/` and the css sheet in `docs/_static/`, which I guess I could also move into the main docs folder.

OrangeChannel:
If you can get a github action working for this to do exactly that—building this to a gh-pages branch and just getting rid of the html folder—I'll change it so the css file is in the main folder and I can re-organize some of the commits here.

kageru:
I thought the documentation was something that could be built by GH actions and then hosted externally, not that we’d have to add thousands of lines of html/js and a lot of fonts to the repository and a makefile (can those please die?) plus a… batch script?
This PR essentially converts us from 50% docstrings and 50% actual code to 5% docstrings, 5% code, and 90% boilerplate to pretty-print our docstrings.

OrangeChannel:
It could be hosted on [RTD](https://docs.readthedocs.io/en/stable/) which is what lvsfunc does, but @stuxcrystal said
> I suggest using proper GitHub Pages.

Which I don't believe can work with externally hosted files, we'd have to build this to a branch's docs folder here. I have it on my master branch now because my gh-pages was only letting me pick master/docs folder as the source.

kageru:
@stuxcrystal any reasons for using github pages that would justify all the bloat in the repository? Otherwise I’d be inclined to agree with hosting on RTD.

stuxcrystal:
No. No issues there.

kageru:
@OrangeChannel could you change the PR so we only generate the docs with github actions and then host them on RTD (i.e. what you did for lvsfunc)? Stux offered `vsutil.encode.moe` as a custom domain for that.

OrangeChannel:
louis did the RTD for lvsfunc but I should be able to figure it out, and I don't think it requires an actions on github's side, just some setup on rtd's side, <https://readthedocs.org/projects/lvsfunc/builds/11507818/>. I can hopefully finish it tonight and add you guys as contributors, if not by tomorrow night.

FichteFoll:
See [this repo](https://github.com/SublimeText/sublime_lib/) for reference, where I set up a github action to build docs from the source tree and push the build results to the `gh-pages` branch. Using RTD and github pages are mutually exclusive; you'd only do either.

OrangeChannel:
@stuxcrystal @kageru alright, all ready to go now I believe. Live version hosted at <http://vsutil.encode.moe/en/latest/>. I've added stux as a contributor to the project <https://readthedocs.org/dashboard/vsutil/users/> but I can't add kageru since you haven't linked your GitHub account with RTD. Once merged, I'll switch the host repo/branch to this one's master branch. Though, @stuxcrystal might have to change it because it creates a webhook on this repo to auto-build on any push, and not sure I have the permissions to do that.

OrangeChannel:
Okay, good to go now, CSS sheet gets loaded correctly.

docs/conf.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+# -- Project information -----------------------------------------------------
+
+project = 'vsutil'
+copyright = '2020, Irrational Encoding Wizardry'
+author = 'Irrational Encoding Wizardry'
+
+# The full version, including alpha/beta/rc tags
+version = release = '0.5.0'
```
kageru, line 26: Is this going to be another place I need to remember updating whenever we release a new version?

docs/_static/css/theme_overrides.css
```diff
@@ -0,0 +1,19 @@
...
+    font-family: SFMono-Regular,Menlo,Monaco,Consolas,"Liberation Mono","Courier New",monospace
+}
+
+.rst-content code.literal {
+    color: #595959
+}
+
+.rst-content .viewcode-back,.rst-content .viewcode-link,.rst-content a code.xref {
+    color: #007020
+}
```
kageru, line 19: I’d prefer to just keep the default RTD theme unless there’s a good reason to change it. These changes seem rather subjective for the most part.

vsutil/clips.py
```diff
@@ -22,20 +22,34 @@ def depth(clip: vs.VideoNode,
...
+    >>> src_10.format.name
+    'YUV420P10'
+
+    >>> src2_10 = vs.core.std.BlankClip(format=vs.RGB30)
+    >>> src2_8 = depth(src2_10, 8, dither_type=Dither.RANDOM)  # override default dither behavior
+    >>> src2_8.format.name
+    'RGB24'
+
+    :param clip:         Input clip.
+    :param bitdepth:     Desired `bits_per_sample` of output clip.
```
kageru, line 39: Are the variable names supposed to be in single backticks? You do that in a lot of places (`bits_per_sample` here, `sample_type` and `clip` later, `frame` further down, but it doesn’t feel consistent to me.

vsutil/func.py
```diff
@@ -1,5 +1,5 @@
 """
-_
+Non-vapoursynth related functions and decorators.
```
kageru, line 2: Since both of our current decorators are for checking clip properties, I’d argue that “non-vapoursynth” isn’t really accurate.

vsutil/types.py
```diff
@@ -12,20 +12,26 @@
 
 class Dither(str, Enum):
     """
-    enum for zimg_dither_type_e
+    Enumeration for `zimg_dither_type_e`.
```
kageru, line 15: I get that “Enumeration” is probably the correct term, but I’ve never met anyone who doesn’t just say enum, so this might cause more confusion than it prevents.

vsutil/types.py
```diff
@@ -12,20 +12,26 @@
...
+    RANDOM = 'random'
+    """Pseudo-random noise of magnitude 0.5."""
+    ERROR_DIFFUSION = 'error_diffusion'
+    """Floyd-Steinberg error diffusion."""
 
 
 class Range(int, Enum):
     """
-    enum for zimg_pixel_range_e
+    Enumeration for `zimg_pixel_range_e`.
```
kageru, line 29: same here

docs/api.rst
```diff
@@ -0,0 +1,58 @@
+=============
+API Reference
+=============
+
+.. automodule:: vsutil
+
+
+Functions that return a clip
+============================
```
kageru, line 9: We might want to think about a check that each new public function is also added here (because I feel like I’m going to forget that at some point), but that’s probably out of scope for this PR. Should be doable with just `grep` alone.

docs/conf.py
stuxcrystal: What about bumpversion?

FichteFoll: It's interpreted Python code, so you could read/load it from somewhere.

OrangeChannel: So, I tried to do this all with an import from a `__version__` dunder in the main module, (it would be here and in setup.py and for anywhere else we needed it in the future) but it doesn't work, the package gets like halfway initialized and then says something like unknown relative import with no parent package or something like that. Unless I was doing something wrong, I think sadly yes, we'd now have two spots where we have to bump the version.

docs/_static/css/theme_overrides.css
OrangeChannel: The colors of the non-stdlib things being red is super obnoxious (i.e. things like `None` and `True` being bright red everywhere on the page) and the stuff that is actually clickable (cross-refs) is not made apparent so I changed it to the same color green as is in the `[source]` plug-in that comes with Sphinx. See <https://lvsfunc.readthedocs.io/en/latest/#lvsfunc.scale.test_descale> for cross-refs and things like `int` just being the same color although you can click on only the cross-refs. And see <https://lvsfunc.readthedocs.io/en/latest/#lvsfunc.scale.test_descale> for the red text, which btw appears a lot more on vsutil's docs than it does here.

The max-width is something stux brought up in Discord way back when the autosummary stuff was being hidden with a scroll-bar, and 1000 seems to be a happy medium between the default 800px and the max on a normal 16:9 1080p screen, which would be like 1600-1700-ish.

I'd be willing to get rid of the width thing, and maybe the font stack, but the colors of the default theme for in-line code is seriously terrible and lacking.

vsutil/clips.py
OrangeChannel: Makes it a bit clearer on the page when I'm refering to "a clip" vs. the input parameter _clip_, as single backticks just make them italics in rst. Subjective since there's no real style guide here, but I'm pretty sure I was consistent with how I did it. I always avoided referencing the parameters in the first line, or the one-sentence docstring summary, but would use them anywhere else. Here, `bits_per_sample` actually refers to the `vapoursynth.Format` parameter, not something from the function, so this is a bit different, but felt this didn't belong in in-line code either.

vsutil/func.py
OrangeChannel: Hmm, yea was wondering how to word this, the non-vapoursynth here applies to the functions, the decorators are the only thing on the file that use the vapoursynth module. What would be a better name here you think?

docs/api.rst
OrangeChannel: Well, this would be doable with Sphinx via the auto-module, but the way we export everything from the submodules directly means this isn't really feasible without the order being weird, and there being prefixes that I'm not sure we even want to encourage as the structure could still change. i.e. the automodule will say `vsutil.info.get_w` instead of `vsutil.get_w` (I can double check this again but pretty sure).

docs/_static/css/theme_overrides.css
OrangeChannel: Default:
image
Custom CSS:
image

docs/conf.py
stuxcrystal: So you don't have to google: https://pypi.org/project/bump2version/

This project can automate bumping the version strings as they tend to be repeated at a lot of places.

docs/_static/css/theme_overrides.css
kageru: > the stuff that is actually clickable (cross-refs) is not made apparent

Has that ever been brought up in an issue upstream? Changing this sounds like something most users would benefit from.

I’d like to get rid of the font overrides (because that’s really just personal preference), but I see your point for the rest. Good changes.

docs/conf.py
kageru: Seems overengineered to me. `sed 's/0\.5\.0/0.6.0/g` should work just as well.

vsutil/clips.py
kageru: I don’t have strong feelings about this, but I don’t think we’ll be able to keep it consistent. My intuition would have been to put all variable/parameter names in `code`, but it’s probably just a waste of time to think too hard about when to italicize something in docstrings. We’ll keep it like this.

vsutil/func.py
kageru: Maybe just change the order to make it clear that non-vapoursynth only refers to the functions.
“Decorators and non-vapoursynth-related (should that be double hyphenated?) functions”

docs/api.rst
kageru: Then automodule might not be the right tool here. If this ever becomes an issue, I’ll just write a small shell script that greps through the source files and the docs and add that to the CI to make sure we didn’t forget anything.

docs/conf.py
FichteFoll: Or a python regexp, but bump2version seems like a useful tool regardless, especially if there is already more than one location where it needs to be updated.

vsutil/func.py
OrangeChannel: Not sure about the double hyphen but I think the point gets across either way.

OrangeChannel:
Any ETA on this @kageru ?

OrangeChannel:
Changed the project URL to this repo and forced a re-build: https://readthedocs.org/projects/vsutil/builds/11657072/

Let me know if you make an RTD account so I can add you to the maintainers list.

PR #47, opened 2020-08-05
split for a single-plane-clip is a noop.

stuxcrystal:


kgrabs:
How about doing the same for `join`? Should be useful for per-plane processing, with how many filters only allow one value per call.

```py
def BoxBlur(clip, hradius=[6,3,3], vradius=[6,3,3], planes=None):
    planes = list(range(clip.format.num_planes) if planes is None else [planes] if isinstance(planes, int) else planes
    clip = split(clip)
    for x in planes:
        clip[x] = core.std.BoxBlur(clip[x], hradius=hradius[x], vradius=vradius[x])
    return join(clip)
```

stuxcrystal:
I added the changes kgrabs requested with additional checks to ensure that the join function continues to behave as expected.

As I result we should think about giving more examples to the documentation to avoid unexpected errors during per-plane processing:

```py
def BoxBlur(clip, hradius=[6,3,3], vradius=[6,3,3], planes=None):
    return join([
        plane.std.BoxBlur(hradius=hradius[pi], vradius=vradius[pi]
        for pi, plane in enumerate( split(clip))
    ], clip.format.color_family)
```

Notice that I pass the color_family manually as `join` will just assume you passed a YUV clip.

vsutil/clips.py
```diff
@@ -113,8 +113,22 @@ def insert_clip(clip: vs.VideoNode, /, insert: vs.VideoNode, start_frame: int) -
...
+        # Equivalent to planes[0] but is supported by more types.
+        plane_clip = next(iter(planes))
+        if family != vs.GRAY:
+            raise ValueError('If passing only a single plane to join, the output family must be vs.GRAY')
+
+        # This should probably be an error to use join like this, but we let it pass.
+        if plane_clip.format.num_planes != 1:
+            plane_clip = plane(plane_clip, 0)
+        
+        return plane_clip
```
OrangeChannel, line 131: This seems really over-engineered, when really all that has to be done is  making a single-item join a no-op by returning *planes[0]*

The function could be simply:

```py
def join(planes: List[vs.VideoNode], family: vs.ColorFamily = vs.YUV) -> vs.VideoNode:
    for plane in planes:
        if plane.format.num_planes > 1:
            raise ValueError("join: all planes should be single-plane grayscale VideoNodes")
    if len(planes) not in [1, 3]:
      raise ValueError("join: planes argument must be exactly 1 or 3 items long")
      
    return planes[0] if len(planes) == 1 else core.std.ShufflePlanes(planes, planes=[0, 0, 0], colorfamily=family)
```

and arguably, we don't really need to check if the clips are single-plane, so it could be as simple as

```py
def join(planes: List[vs.VideoNode], family: vs.ColorFamily = vs.YUV) -> vs.VideoNode:
    if len(planes) not in [1, 3]:
      raise ValueError("join: planes argument must be exactly 1 or 3 items long")

    return planes[0] if len(planes) == 1 else core.std.ShufflePlanes(planes, planes=[0, 0, 0], colorfamily=family)
```

The check for number of planes given though has to stay, since `core.std.ShufflePlanes([Y,U], planes=[0,0,0], colorfamily=vs.YUV)` is a valid (but not desirable) call.

kageru:
closing in favor of #61

PR #48, opened 2020-08-09
Implement doodle1 SplitPlanes

stuxcrystal:


vsutil/clips.py
```diff
@@ -131,6 +131,13 @@ def plane(clip: vs.VideoNode, planeno: int, /) -> vs.VideoNode:
     """
     if clip.format.num_planes == 1 and planeno == 0:
         return clip
+
+    if hasattr(clip.std, "SplitPlanes"):
+        planes = clip.std.SplitPlanes()
+        if isinstance(planes, vs.VideoNode):
+            planes = [planes]
+        return planes[planeno]
```
OrangeChannel, line 139:
```suggestion
    if hasattr(clip.std, "SplitPlanes"):
        planes = clip.std.SplitPlanes()
        return planes if isinstance(planes, vs.VideoNode) else planes[planeno]
```
This can be simplified but actually, if I understand what `SplitPlanes` does, we don't even need to catch this here since we already have a no-op on single-plane clips. (L132-133)

@stuxcrystal when would there be a case that `clip.format.num_planes == 1 and planeno == 0` would be false while SplitPlanes only returns a single plane?

If not:

```suggestion
    if hasattr(clip.std, "SplitPlanes"):
        return clip.std.SplitPlanes()[planeno]
```

vsutil/clips.py
```diff
@@ -139,6 +146,11 @@ def split(clip: vs.VideoNode, /) -> List[vs.VideoNode]:
     """
     Returns a list of planes for the given input clip.
     """
+    if hasattr(clip.std, "SplitPlanes"):
+        planes = clip.std.SplitPlanes()
+        if isinstance(planes, vs.VideoNode):
+            planes = [planes]
+        return planes
```
OrangeChannel, line 153: I think the only reason to keep this section here is because it might be nanoseconds faster to call SplitPlanes once instead of three times for a 3-plane clip, but it does exactly what my updated version of `plane` would return in the current return statement for `split` (L154). I guess we could keep this if a call to `SplitPlanes` with a 1-plane clip is truly a no-op (it seems to be from the [code in simplefilters.cpp](https://github.com/vapoursynth/vapoursynth/commit/af5147659034c1d8641f9039d213b722e10e0d72#diff-47d03cdc5bf112da0bb1b119d542907dR689)) but if it's not, then this actually breaks the current behavior of returning the same object just in a one item long list.

Similar to the other comment,
```suggestion
    if hasattr(clip.std, "SplitPlanes"):
        planes = clip.std.SplitPlanes()
        return [planes] if isinstance(planes, vs.VideoNode) else planes
```

OrangeChannel:
We might also want to add to the documentation that `split` shouldn't really be used over `core.std.SplitPlanes`, we won't deprecate it since it has slighly different behavior (i.e. always returning a list) but this could be done in-line with `SplitPlanes` once doodle1 is merged.

vsutil/clips.py
stuxcrystal: I did not add the single-plane check because of https://github.com/Irrational-Encoding-Wizardry/vsutil/pull/47

OrangeChannel: See my comment: https://github.com/Irrational-Encoding-Wizardry/vsutil/pull/47#discussion_r469753129

`split` already is a no-op via `plane` (depends on your definition of no-op I guess but it returns the same object) for single-plane clips, this was changed a few months ago.

kageru:
closing in favor of #61

PR #50, opened 2020-08-13
Better enum repr

OrangeChannel:
Old behavior:

```py
>>> vsutil.types._readable_enums(Range)
'<vsutil.types.Range.LIMITED: 0>, <vsutil.types.Range.FULL: 1>'

>>> vsutil.types._resolve_enum(Dither, 'bad', 'test')
ValueError: test must be in <vsutil.types.Dither.NONE: none>, <vsutil.types.Dither.ORDERED: ordered>,
<vsutil.types.Dither.RANDOM: random>, <vsutil.types.Dither.ERROR_DIFFUSION: error_diffusion>.

>>> vsutil.types._readable_enums(vs.SampleType)
'<importlib._bootstrap.SampleType.INTEGER: 0>, <importlib._bootstrap.SampleType.FLOAT: 1>'

>>> vsutil.types._readable_enums(vs.SampleType, 'vapoursynth')  # workaround
'<vapoursynth.SampleType.INTEGER: 0>, <vapoursynth.SampleType.FLOAT: 1>'
```

New behavior:
```py
>>> vsutil.types._readable_enums(Range)
'<vsutil.Range.LIMITED: 0>, <vsutil.Range.FULL: 1>'

>>> vsutil.types._resolve_enum(Dither, 'bad', 'test')
ValueError: test must be in <vsutil.Dither.NONE: 'none'>, <vsutil.Dither.ORDERED: 'ordered'>,
<vsutil.Dither.RANDOM: 'random'>, <vsutil.Dither.ERROR_DIFFUSION: 'error_diffusion'>.

>>> vsutil.types._readable_enums(vs.SampleType)
'<vapoursynth.INTEGER: 0>, <vapoursynth.FLOAT: 1>'
```

Basically, since we re-export everything at the main module level, the default repr of
`'<%s.%s.%s: %s>' % (self.__module__.__name__, self.__class__.__name__, self.name, self.value)`
includes the submodule name, which we probably don't want. The old way of getting the vapoursynth enums to work was also a hack, but an uglier hack than the one I added here (it also doesn't print the enum name anymore for vapoursynth as vs exports the enum items directly).

This restores the behavior from https://github.com/Irrational-Encoding-Wizardry/vsutil/commit/9de51192b77b62d733d8b541af912a5a30dfee13 that was lost when we split the package into submodules.

Also simplifies the `vapoursynth` module name override logic originally introduced in https://github.com/Irrational-Encoding-Wizardry/vsutil/commit/424102be24dac8bdebd0b77201ec889758133efc.

### DIFF
--- a/vsutil/types.py
+++ b/vsutil/types.py
@@ -5,7 +5,7 @@ __all__ = ['Dither', 'Range']
 
 # this file only depends on the stdlib and should stay that way
 from enum import Enum
-from typing import Any, Optional, Type, TypeVar, Union
+from typing import Any, Callable, Optional, Type, TypeVar, Union
 
 E = TypeVar('E', bound=Enum)
 
@@ -45,7 +45,7 @@ def _readable_enums(enum: Type[Enum]) -> str:
         return ', '.join([repr(i) for i in enum])
 
 
-def _resolve_enum(enum: Type[E], value: Any, var_name: str, module: Optional[str] = None) -> Union[E, None]:
+def _resolve_enum(enum: Type[E], value: Any, var_name: str, fn: Optional[Callable] = None) -> Union[E, None]:
     """
     Attempts to evaluate `value` in `enum` if value is not None, otherwise returns None.
     Basically checks if a supplied enum value is valid and returns a readable error message
@@ -56,4 +56,4 @@ def _resolve_enum(enum: Type[E], value: Any, var_name: str, module: Optional[str
     try:
         return enum(value)
     except ValueError:
-        raise ValueError(f'{var_name} must be in {_readable_enums(enum, module)}.') from None
+        raise ValueError(f"{fn.__name__ + ': ' if fn else ''}{var_name} must be in {_readable_enums(enum)}.") from None
--- a/vsutil/types.py
+++ b/vsutil/types.py
@@ -10,7 +10,13 @@ from typing import Any, Optional, Type, TypeVar, Union
 E = TypeVar('E', bound=Enum)
 
 
-class Dither(str, Enum):
+class _NoSubmoduleRepr(Enum):
+    def __repr__(self):
+        """Removes submodule name from standard repr, helpful since we re-export everything at the top-level."""
+        return '<%s.%s.%s: %r>' % (self.__module__.split('.')[0], self.__class__.__name__, self.name, self.value)
+
+
+class Dither(str, _NoSubmoduleRepr):
     """
     enum for zimg_dither_type_e
     """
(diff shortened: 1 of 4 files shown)


vsutil/types.py
```diff
@@ -47,4 +56,4 @@ def _resolve_enum(enum: Type[E], value: Any, var_name: str, module: Optional[str
     try:
         return enum(value)
     except ValueError:
-        raise ValueError(f'{var_name} must be in {_readable_enums(enum, module)}.') from None
+        raise ValueError(f"{fn.__name__ + ': ' if fn else ''}{var_name} must be in {_readable_enums(enum)}.") from None
```
FichteFoll, line 59: The function will already be part of the traceback printed on the console (albeit on the second-to-last frame). Do we really need to pass the calling function to mention the function name again?

vsutil/types.py
```diff
@@ -5,12 +5,18 @@
...
 # this file only depends on the stdlib and should stay that way
 from enum import Enum
-from typing import Any, Optional, Type, TypeVar, Union
+from typing import Any, Callable, Optional, Type, TypeVar, Union
 
 E = TypeVar('E', bound=Enum)
 
 
-class Dither(str, Enum):
+class _NoSubmoduleRepr(Enum):
```
FichteFoll, line 13: I would define `_NoSubmoduleRepr` as a mixin instead of subclassing Enum with it, i.e. `class Range(int, _NoSubmoduleRepr, Enum):`.

vsutil/types.py
OrangeChannel: I'd like to keep the standard of `raise Error("func_name: the problem") because this will be used in scripts, which will likely be ran through editors like VSEdit etc., where the person running them might have a hundred other filters and function calls with similar named variables or stuff. The Python tracebacks should make it clear where the error happened but for people new to vs (who likely don't know anything about Python nor what to do with the traceback besides posting it on discord and just saying "help"), printing the error message with the function name that caused the issue is probably a bit more clear IMO.

vsutil/types.py
OrangeChannel: Good idea, changed in https://github.com/Irrational-Encoding-Wizardry/vsutil/pull/50/commits/80a21103365fd9f0e303ac744a075a71ee65be6f with a slight difference. The mixin class MUST come first otherwise `int` and `str`'s `__repr__` overrides our custom one, you can see that here: https://nbviewer.jupyter.org/gist/OrangeChannel/49a58bb6af63fc1baddd30ed68574479

PR #51, opened 2020-08-25
Create a _metadata.py file to re-use version number and author

OrangeChannel:
This makes it so the version number only needs to be bumped in one file instead of the docs/conf.py, setup.py, and \_\_init\_\_.py.

Can be modified in the future if we want to add more fields like a date. There's also the possibility to auto-update the file with git tags via git describe etc. but probably not needed for now.

Not sure we need the try/except block in \_\_init\_\_.py but if someone/something packages this without the metadata file for whatever reason, would be nice to have the author and version dunders still defined.

The only thing this changes externally is it adds a `vsutil.__version__` and `vsutil.__author__` which are really only used by documentation services/scripts like pydoc but might also help with people reporting errors, since we can now ask them to tell us what `>>> print(vsutil.__version__)` gives.

setup.py and the docs/conf.py should never need to be modified again unless the domain name changes or if we decide to host on something besides github.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,19 @@
 from setuptools import setup, find_packages
 from setuptools.command.test import test
+from distutils.util import convert_path
+
+# We can't import the submodule normally as that would "run" the main module
+# code while the setup script is meant to *build* the module.
+
+# Besides preventing a whole possible mess of issues with an un-built package,
+# this also prevents the vapoursynth import which breaks the docs on RTD.
+
+# convert_path is used here because according to the distutils docs:
+# '...filenames in the setup script are always supplied in Unix
+# style, and have to be converted to the local convention before we can
+# actually use them in the filesystem.'
+meta = {}
+exec(open(convert_path('vsutil/_metadata.py')).read(), meta)
 
 
 class DiscoverTest(test):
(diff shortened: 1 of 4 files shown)


vsutil/__init__.py
```diff
@@ -14,3 +14,10 @@
 __all__ = []
 for _pkg in _mods:
     __all__ += __import__(__name__ + '.' + _pkg, fromlist=_mods).__all__
+
+try:
+    from . import _metadata
+    __author__ = _metadata.__author__
+    __version__ = _metadata.__version__
```
FichteFoll, line 21:
```suggestion
    from ._metadata import __author__, __version__
```

setup.py
```diff
@@ -1,5 +1,9 @@
 from setuptools import setup, find_packages
 from setuptools.command.test import test
+from distutils.util import convert_path
+
+meta = {}
+exec(open(convert_path('vsutil/_metadata.py')).read(), meta)
```
FichteFoll, line 6: I suppose the reason you're not importing it normally is so it doesn't require vapoursynth? If so, I believe that deserves a comment.

FichteFoll (commented):
I think this is a good solution without resorting to external tools like bump2version.

setup.py
OrangeChannel: https://github.com/Irrational-Encoding-Wizardry/vsutil/pull/51/commits/5d7689d039c12e4341325d3d622056e487886737#diff-2eeaed663bd0d25b7e608891384b7298R5-R14

If the module was a single file, then yes, RTD not being able to build vapoursynth would be the reason to not import normally. But, it gets a bit more complicated when it's a package with multiple relative imports in it, that at the start of this script is uninitialized or "partially initialized" which causes circular import issues and the issue we have with `types.py`'s namespace.
I also believe this breaks installing it with pep as it will attempt to run the uninitialized module before knowing what dependencies are needed (in this case just vapoursynth).

FichteFoll: Yes, it would fail for setup.py and for RTD, neither of which would have `vapoursynth` importable. The comment might be a bit verbose, but that's better than no comment imo.

OrangeChannel: The error message from the python console in pycharm mentions circular imports because of our types submodule trying to import from enum which imports from the stdlib types. Since the package isn't built, stdlib enum tries to import from our types instead as it doesn't know it's namespace is actually vsutil.types.

PR #55, opened 2021-05-25
Add get_sample_type function

Ichunjo:
I currently use [internally](https://github.com/Ichunjo/vardefunc/blob/e20a8c48149e34f5927b9d9f67159ee0357b498d/vardefunc/util.py#L31#L35), but I thought it might be a good addition to vsutil.

### DIFF
--- a/vsutil/info.py
+++ b/vsutil/info.py
@@ -61,6 +61,21 @@ def get_plane_size(frame: Union[vs.VideoFrame, vs.VideoNode], /, planeno: int) -
 
 
 @func.disallow_variable_format
+def get_sample_type(clip: vs.VideoNode, /) -> vs.SampleType:
+    """Returns the sample type of a VideoNode as an SampleType.
+
+    >>> src = vs.core.std.BlankClip(format=vs.YUV420P16)
+    >>> get_sample_type(src)
+    <SampleType.INTEGER: 0>
+
+    :param clip:  Input clip.
+
+    :return:      Sample type of the input `clip`.
+    """
+    return clip.format.sample_type
+
+
+@func.disallow_variable_format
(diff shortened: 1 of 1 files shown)


vsutil/info.py
```diff
@@ -60,6 +60,21 @@ def get_plane_size(frame: Union[vs.VideoFrame, vs.VideoNode], /, planeno: int) -
...
+
+    >>> src = vs.core.std.BlankClip(format=vs.YUV420P16)
+    >>> get_sample_type(src)
+    <SampleType.INTEGER: 0>
+
+    :param clip:  Input clip.
+
+    :return:      Sample type of the input `clip`.
+    """
+    return clip.format.sample_type
```
FichteFoll, line 75: I have to assume the sole purpose for this function is to make use of the `disallow_variable_format` decorator? Because I don't see the advantage in `get_sample_type(clip)` over `clip.format.sample_type`.

Ichunjo: `get_depth` uses the same logic though, so I didn't think it was necessary.

kageru: `get_depth` is a very common operation, so saving some typing there seemed like a good idea at the time, with the additional benefit of catching breaking changes in the Vapoursynth API.
I don’t think `get_sample_type` would see the same kind of usage tbh.

FichteFoll: > `get_depth` uses the same logic though, so I didn't think it was necessary.

Well, I'm with you on that being borderline, but the main reason for why I consider `get_depth` to be warranted is that it uses a different name for this info as opposed to `clip.format.bits_per_sample`. Now, I personally would be fine with the latter, but I at least see some use of having that alias.

PR #60, opened 2021-10-23
Deprecate split function in favor of std.SplitPlanes

kageru:


kageru:
closing in favor of #61

PR #61, opened 2021-12-28
Update join/split

Ichunjo:


vsutil/clips.py
```diff
@@ -138,20 +138,19 @@ def insert_clip(clip: vs.VideoNode, /, insert: vs.VideoNode, start_frame: int) -
     return pre + insert + post
 
 
-def join(planes: List[vs.VideoNode], family: vs.ColorFamily = vs.YUV) -> vs.VideoNode:
+def join(planes: Sequence[vs.VideoNode], family: vs.ColorFamily = vs.YUV) -> vs.VideoNode:
     """Joins the supplied list of planes into a three-plane VideoNode (defaults to YUV).
 
     >>> planes = [Y, U, V]
     >>> clip = join(planes)
```
OrangeChannel, line 145:
```suggestion
    """Joins the supplied sequence of planes into a single VideoNode (defaults to YUV).

    >>> planes = [Y, U, V]
    >>> clip_YUV = join(planes)
    >>> plane = core.std.BlankClip(format=vs.GRAY8)
    >>> clip_GRAY = join([plane], family=vs.GRAY)
```

vsutil/clips.py
```diff
@@ -138,20 +138,19 @@ def insert_clip(clip: vs.VideoNode, /, insert: vs.VideoNode, start_frame: int) -
...
 
-    :param planes:  List of 3 one-plane ``vapoursynth.GRAY`` clips to merge.
+    :param planes:  Sequence of 3 one-plane ``vapoursynth.GRAY`` clips to merge.
     :param family:  Output color family. Must be a three-plane color family.
 
     :return:        Three-plane clip of the supplied `planes`.
```
OrangeChannel, line 150:
```suggestion
    :param planes:  Sequence of one-plane ``vapoursynth.GRAY`` clips to merge.
    :param family:  Output color family.

    :return:        Merged clip of the supplied `planes`.
```

vsutil/clips.py
```diff
@@ -178,13 +177,13 @@ def split(clip: vs.VideoNode, /) -> List[vs.VideoNode]:
     """Returns a list of planes (VideoNodes) from the given input clip.
 
     >>> src = vs.core.std.BlankClip(format=vs.RGB27)
-    >>> R, G, B  = split(src)
+    >>> R, G, B = split(src)
```
OrangeChannel, line 181:
```suggestion
>>> src2 = vs.core.std.BlankClip(format=vs.GRAY8)
>>> split(src2)
[<vapoursynth.VideoNode object>]  # always returns a list, even if single plane

```

OrangeChannel: Maybe there's a better way to show this...

OrangeChannel (changes requested):
Some doc changes and examples.

vsutil/clips.py
```diff
@@ -138,20 +138,21 @@ def insert_clip(clip: vs.VideoNode, /, insert: vs.VideoNode, start_frame: int) -
...
-    :param planes:  List of 3 one-plane ``vapoursynth.GRAY`` clips to merge.
-    :param family:  Output color family. Must be a three-plane color family.
+    :param planes:  Sequence of one-plane ``vapoursynth.GRAY`` clips to merge.
+    :param family:  Output color family.
 
-    :return:        Three-plane clip of the supplied `planes`.
+    :return:        Merged clip of the supplied `planes`.
     """
-    if family not in [vs.RGB, vs.YUV]:
-        raise ValueError('Color family must have 3 planes.')
```
FichteFoll, line 153: Maybe some kind of look-up table for the amount of planes of a format would be helpful here? I'm not aware of amounts other than 1 or 3, but I also don't have the full picture. Since the restriction has been removed in the docstring, I'd like to make sure that it is truly generic over any format.

FichteFoll: Actually, you are already providing a sequence whose length is known to the function, so one could already do `core.std.ShufflePlanes(planes, [0] * len(planes), family)` and have vs throw an exception if the family isn't compatible with the number of planes. Assuming it does that.

kageru: It doesn’t. If I try to `ShufflePlanes` two clips into YUV, it will take the second clip as both the U and V plane.

PR #64, opened 2022-02-25
Fix plane behaviour with RGB clips

Setsugennoao:


kageru:
Could you explain the motivation behind this change? I see a lot of added complexity for seemingly little benefit.

kageru:
Please add a test that shows RGB inputs no longer raising errors.

PR #65, opened 2022-02-25
Update iterate and add kwargs param

Setsugennoao:


vsutil/func.py
```diff
@@ -64,7 +64,7 @@ def fallback(value: Optional[T], fallback_value: T) -> T:
     return fallback_value if value is None else value
 
 
-def iterate(base: T, function: Callable[[Union[T, R]], R], count: int) -> Union[T, R]:
+def iterate(base: T, function: Callable[[Union[T, R]], R], count: int, **kwargs: Any) -> Union[T, R]:
```
kageru, line 67: I have a feeling this will no longer pass typechecks because this `Callable` doesn’t know it accepts any kwargs.
I also don’t know if there is any way to make this work other than just removing the type hints altogether, but how necessary is this change anyway? You could have done the functools.partial at call site.

PR #66, opened 2022-02-25
Update disallow_variable_* functions

Setsugennoao:


kageru:
Could you explain the motivation behind this change? I see a lot of added complexity for seemingly little benefit.

PR #70, opened 2022-06-16
ci: Use nix for tests on darwin and linux

cid-chan:
Also: Tests run with r59.

If darwin fails: It didn't pick a runner for 20 minutes and I need to sleep.


flake.nix
```diff
@@ -0,0 +1,60 @@
...
+          (python.withPackages (ps: [
+            vapoursynth_python
+          ]))
+        ];
+      };
+      devShell = self.devShell.${system}.default;
+
+      packages.default = python.pkgs.buildPythonPackage {
+        pname = "vsutil";
+        version = "0.7.0";
```
kageru, line 48: Will we have to change this every time we release a new version?

cid-chan: It needs a version. I just provided one. I don't really care if it's up to date.

PR #72, opened 2022-06-21
Fix Sphinx builds

RivenSkaye:
Bumps a couple versions in `docs/requirements.txt` and bumps the Python version used in `.readthedocs.yml`.
Sphinx builds were tested and checked from a venv to ensure reproducible results.
For completeness sake, the things installed on this venv are:
- `vapoursynth-portable`
- `vsutil (git master)`
- anything `pip install -r docs/requirements.txt` installed and its dependencies.

Optional:
- Bump ` docutils`  to 0.18, as Sphinx lists the versions allowed as `<0.19,>=0.16`
- Remove `Pygments` or ease the version pinned, as Sphinx should require a version itself

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,7 +1,7 @@
 version: 2
 
 python:
-   version: 3.10
+   version: 3.8
    install:
        - requirements: docs/requirements.txt
 
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,7 +1,7 @@
 version: 2
 
 python:
-   version: 3.8
+   version: 3.10
    install:
        - requirements: docs/requirements.txt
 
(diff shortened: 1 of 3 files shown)


RivenSkaye:
This PR can not be merged yet thanks to RTD breaking on builds. Seems to be an issue with Sphinx requiring `jinja2` and then failing to resolve an import

RivenSkaye:
That last commit fixed it, view the build [on RTD](https://vsutil-riven.readthedocs.io/en/latest/api.html) and notify me here or through Discord if there are any issues

RivenSkaye:
We could, but AFAIK that would require a git runner to build the HTML docs and restructuring the current layout to accommodate the way GH Pages builds with jekyll, or passes it through if no info is found. And git runners have been becoming an issue lately.
Another alternative could be to have a pre-commit hook that prevents pushing/merging to master without updating the docs, but I can see that cause a lot of frustration for devs that don't want to run Sphinx locally

PR #74, opened 2022-07-08
Add shift_clip function

Setsugennoao:
Original taken from https://github.com/kgrabs/zzfunc/blob/master/zzfunc/std.py#L211

### DIFF
--- a/vsutil/clips.py
+++ b/vsutil/clips.py
@@ -1,7 +1,15 @@
 """
 Functions that modify/return a clip.
 """
-__all__ = ['depth', 'frame2clip', 'get_y', 'insert_clip', 'join', 'plane', 'split']
+__all__ = [
+    'depth',
+    # frame util
+    'frame2clip',
+    # plane utils
+    'get_y', 'join', 'plane', 'split',
+    # clip splicing utils
+    'insert_clip', 'shift_clip', 'shift_clip_multi'
+]
 
 from typing import Any, List, Optional, Sequence, Union, cast
 
@@ -229,3 +237,21 @@ def _should_dither(in_bits: int,
                 or float_to_int
                 or (in_range == types.Range.FULL and upsampling and (in_bits, out_bits) != (8, 16))
                 or downsampling)
+
+
+def shift_clip(clip: vs.VideoNode, offset: int) -> vs.VideoNode:
+    if offset > clip.num_frames - 1:
+        raise ValueError("shift_clip: Offset can't be greater than the clip length!")
+
+    if offset < 0:
+        return clip[0] * abs(offset) + clip[:offset]
+    elif offset > 0:
+        return clip[offset:] + clip[-1] * offset
+
+    return clip
+
+
+def shift_clip_multi(clip, shift: types.FrameRange = (-1, 1)):
+    ranges = func.normalize_franges(shift)
+
+    return [shift_clip(clip, x) for x in ranges]
--- a/vsutil/clips.py
+++ b/vsutil/clips.py
@@ -251,7 +251,7 @@ def shift_clip(clip: vs.VideoNode, offset: int) -> vs.VideoNode:
     return clip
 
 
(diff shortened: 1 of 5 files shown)


OrangeChannel:
Add entry in docs before this is merged.

vsutil/clips.py
```diff
@@ -229,3 +237,21 @@ def _should_dither(in_bits: int,
...
+
+    if offset < 0:
+        return clip[0] * abs(offset) + clip[:offset]
+    elif offset > 0:
+        return clip[offset:] + clip[-1] * offset
+
+    return clip
+
+
+def shift_clip_multi(clip, shift: types.FrameRange = (-1, 1)):
```
FichteFoll, line 254: Missing return type annotation.

(The corresponding mypy option should be set to make incomplete function signatures fail the lint.)

FichteFoll (changes requested):

vsutil/clips.py
Setsugennoao: There is no linting in this package

vsutil/clips.py
```diff
@@ -229,3 +237,21 @@ def _should_dither(in_bits: int,
...
+    elif offset > 0:
+        return clip[offset:] + clip[-1] * offset
+
+    return clip
+
+
+def shift_clip_multi(clip, shift: types.FrameRange = (-1, 1)) -> List[vs.VideoNode]:
+    ranges = func.normalize_franges(shift)
+
+    return [shift_clip(clip, x) for x in ranges]
```
kageru, line 257: Can you give a real-world example for how this would be used, specifically the `_multi` variant?

vsutil/clips.py
```diff
@@ -229,3 +237,21 @@ def _should_dither(in_bits: int,
...
                 or (in_range == types.Range.FULL and upsampling and (in_bits, out_bits) != (8, 16))
                 or downsampling)
+
+
+def shift_clip(clip: vs.VideoNode, offset: int) -> vs.VideoNode:
+    if offset > clip.num_frames - 1:
+        raise ValueError("shift_clip: Offset can't be greater than the clip length!")
+
+    if offset < 0:
+        return clip[0] * abs(offset) + clip[:offset]
```
kageru, line 247: Shouldn’t this be `clip[:-offset]`? Duplicate the first frame `offset` times and then remove the last `offset` frames.

vsutil/func.py
```diff
@@ -142,3 +147,14 @@ def iterate(base: T, function: Callable[[Union[T, R]], R], count: int) -> Union[
...
+
+def normalize_franges(franges: FrameRange, /) -> Iterable[int]:
+    if isinstance(franges, int):
+        return [franges]
+    elif isinstance(franges, tuple):
+        start, stop = franges
+        step = -1 if stop < start else 1
+        return range(start, stop + step, step)
+    else:
+        return franges
```
kageru, line 160: I said before that we’re not doing this, especially not as part of the public API.

kageru (changes requested):

vsutil/clips.py
Setsugennoao: Offset already is negative

vsutil/clips.py
Setsugennoao: Single:
<https://github.com/Setsugennoao/stgfunc/blob/master/stgfunc/mask.py#L162>
Multi:
<https://github.com/Setsugennoao/stgfunc/blob/master/stgfunc/tweaking.py#L292>
<https://github.com/Setsugennoao/Encoding-Scripts/blob/master/Other%20Groups/gsk-kun/Eiga%20Daisuki%20Pompo-san%20(BD)/pompo_san.vpy#L221-L222>

vsutil/func.py
FichteFoll: Can you point me to your exact reasoning? Standardizing a way of frame specification sounds like a fine usage of `vsutil`, provided it also has a way of making the frame specification usable e.g. by returning an iterable over all frame numbers.

PR #76, opened 2022-09-03
func: Add a function `function` that aliases function safely.

cid-chan:
`function(plugin: str, name: str) -> Callable[..., typing.Any]`
Intended use:
```py
class Spline16:
   scaler_function = vsutil.function("resize", "Spline16")
```

Calling the result of function will be equivalent to:
```py
getattr(getattr(vs.core, plugin), name)(*args, **kwargs)
```

This eliminates the risk of doing
```py
class Spline16:
   scaler_function = core.resize.Spline16
```

Which is illegal.

### DIFF
--- a/vsutil/func.py
+++ b/vsutil/func.py
@@ -7,6 +7,8 @@ __all__ = [
     'disallow_variable_format', 'disallow_variable_resolution',
     # misc non-vapoursynth related
     'fallback', 'iterate',
+    # misc vapoursynth related
+    'function'
 ]
 
 import inspect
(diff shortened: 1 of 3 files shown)


FichteFoll:
There is a good working alternative to

```py
class Spline16:
   scaler_function = core.resize.Spline16
```

which is
```py
class Spline16:
   @property
    def scaler_function(self):
        return core.resize.Spline16
```

. How is
```py
class Spline16:
   scaler_function = function("resize", "Spline16")
```
an improvement, using this much wrapper code and more indirection?

stuxcrystal:
Oh, it's part of the VapourSynth API. I thought I could make it compatible so it's less likely to break something when someone switches.

stuxcrystal:
That doesn't work when someone decides to alias at the module level.

FichteFoll:
> That doesn't work when someone decides to alias at the module level.

Is that an actual use case? Does this exist in the wild or is there interest for it? Just asking because this was the criteria you applied to previous additions to vsutil.

vsutil/func.py
```diff
@@ -142,3 +144,61 @@ def iterate(base: T, function: Callable[[Union[T, R]], R], count: int) -> Union[
...
+
+
+# This function is actually implemented as a class.
+# This makes sure that,
+# when it is used as the value of a class-variable,
+# python does not prefix calls to this function with ``self``.
+# It also allows to forward calls to 
+# - `plugin`,
+# - `signature`,
+# - and `return_signature`
```
kageru, line 156: When are `signature` and `return_signature` used by users? I get that they’re part of the VS API and you probably want the return of `function` to be as close to that as possible, but is there a real case of anyone accessing one of these from Python that justifies this layer of indirection (when they could instead call `.resolved` and not rely on us playing the middle man)?

kageru: nvm, that was answered above

kageru (approved):

FichteFoll:
> Is that an actual use case? Does this exist in the wild or is there interest for it? Just asking because this was the criteria you applied to previous additions to vsutil.

A tl;dr for the answer is: better documentation for this vs "quirk" and a way to phase out the workaround in the future compared to copy-pasting the property solution.

PR #78, opened 2022-11-18
Add mod4 output to `get_w`

### LINKED ISSUE #49: get_w
```python
"""
https://github.com/HomeOfVapourSynthEvolution/havsfunc/blob/master/havsfunc.py#L5907
"""
import math


def cround(x):
    return math.floor(x + 0.5) if x > 0 else math.ceil(x - 0.5)


def m4(x):
    return 16 if x < 16 else cround(x / 4) * 4


desc_w = m4(849 * 1920 / 1080)
print(f'havsfunc:   {desc_w}')


"""
https://github.com/Infiziert90/getnative/blob/master/getnative/app.py#L194
"""
def getw(h, only_even=True):
    w = h * 16 / 9
    w = int(round(w))
    if only_even:
        w = w // 2 * 2
    return w


desc_w = getw(849)
print(f'getnative:  {desc_w}')


"""
https://github.com/Irrational-Encoding-Wizardry/kagefunc/blob/master/kagefunc.py#L336
"""
def getw(height, aspect_ratio=16 / 9, only_even=True):
    """
    Returns width for image.
    """
    width = height * aspect_ratio
    width = int(round(width))
    return width // 2 * 2 if only_even else width


desc_w = getw(849)
print(f'kagefunc:   {desc_w}')


"""
https://github.com/Irrational-Encoding-Wizardry/vsutil/blob/master/vsutil/info.py#L74
"""
def get_w(height: int, aspect_ratio: float = 16 / 9, *, only_even: bool = True) -> int:
    """
    Calculates the width for a clip with the given height and aspect ratio.
    only_even is True by default because it imitates the math behind most standard resolutions (e.g. 854x480).
    """
    width = height * aspect_ratio
    if only_even:
        return round(width / 2) * 2
    return round(width)


desc_w = get_w(849)
print(f'vsutil:     {desc_w}')
```
↓
```
havsfunc:   1508
getnative:  1508
kagefunc:   1508
vsutil:     1510
```

RivenSkaye:
Closes #49 by adding a proper way to get the same results as the other func scripts, without changing default behavior for this implementation.
This _can_ actually be useful when working with e.g. interlaced material, as noted by Dave in the comments there.
Also adds a test case for the new param using the (ironically odd) example from the referenced issue.

### DIFF
--- a/tests/test_vsutil.py
+++ b/tests/test_vsutil.py
@@ -162,6 +162,7 @@ class VsUtilTests(unittest.TestCase):
         self.assertEqual(vsutil.get_w(480, only_even=False), 853)
         self.assertEqual(vsutil.get_w(1080, 4 / 3), 1440)
         self.assertEqual(vsutil.get_w(1080), 1920)
+        self.assertEqual(vsutil.get_w(849, mod4=True), 1508)
 
     def test_iterate(self):
         def double_number(x: int) -> int:
--- a/tests/test_vsutil.py
+++ b/tests/test_vsutil.py
@@ -162,7 +162,7 @@ class VsUtilTests(unittest.TestCase):
         self.assertEqual(vsutil.get_w(480, only_even=False), 853)
         self.assertEqual(vsutil.get_w(1080, 4 / 3), 1440)
         self.assertEqual(vsutil.get_w(1080), 1920)
-        self.assertEqual(vsutil.get_w(849, mod4=True), 1508)
+        self.assertEqual(vsutil.get_w(849, mod=4), 1508)
 
     def test_iterate(self):
         def double_number(x: int) -> int:
--- a/vsutil/info.py
+++ b/vsutil/info.py
@@ -94,7 +94,7 @@ def get_subsampling(clip: vs.VideoNode, /) -> Union[None, str]:
         raise ValueError('Unknown subsampling.')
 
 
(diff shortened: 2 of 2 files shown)


LightArrowsEXE:
Is there a reason to limit this to MOD4 only? Why not allow the user to set their own MOD?

RivenSkaye:
That's actually a very valid suggestion, though I fear implementing this in a non-breaking way would result in either keeping an entirely redundant parameter around (`only_even` would do nothing) or changing that parameter's function in a way that would break its described function, as it'd become a flag to use a custom mod, which the user is free to set to any odd value.

The only thing I can come up with that would make sense with the existing parameters in mind is coercing the mod to an even value if the flag is present, defaulting to `mod=1` so that `only_even=False` would provide the same functionality as it does now and so that `only_even=True` would either coerce `ceil(mod / 2) * 2` (`round` would cause this to drop to 0, a blind `mod + 1` would coerce `mod=2` to 4) or coerce it in the same way as it does now, except for the special case where `mod=1`. Either way I'd like to discuss a rather large change of the function before implementing it.
CC @kageru and @cid-chan

kageru:
I I like the suggestions of `mod: int` but with default `None` for now and then do `mod = fallback(mod, 2 if only_even else 1)` and a deprecation warning for `only_even` which keeps the current behavior but allows for arbitrary mods with a path to removing `only_even` without losing any functionality.

